PR #4, opened 2021-11-16
clean up the whole kiso-testing (except jenkins file)

BKaDamien:
Just clean the kiso repo


tests/test_config_parser.py
```diff
@@ -136,9 +136,9 @@ def create_simple_config():
         com:   chan1
         flash: chan2
     config:
```
sebastianpfischer, line 138: I'm a bit confused about why we have here this kind of config, when I take a look at https://github.com/BKaDamien/kiso-testing/blob/cleanup_kiso_testing/src/pykiso/lib/auxiliaries/example_test_auxiliary.py. Let us just delete it or put it to something more generic

sebastianpfischer (changes requested):

sebastianpfischer (approved):

PR #5, opened 2021-11-23
- just add pykiso logo to sphinx documentation

BKaDamien:
Simply add pykiso doc


sebastianpfischer:
Can you also add it into the readme? Like that we can see it directly when we access https://github.com/eclipse/kiso-testing

sebastianpfischer (changes requested):
on the readme.md

sebastianpfischer (approved):

PR #16, opened 2022-02-10
Use timeout to avoid long blocking builds

fredg02:
Please copy the timeout setting to all branches (and adjust if necessary).

Context: I have noticed builds running for 23 hours and just hanging. This unnecessarily binds resources.


sebastianpfischer:
Thanks ^^

sebastianpfischer (approved):

PR #18, opened 2022-02-16
test(test_proxy_auxiliary.py): make test reliable

Pog3k:
multiprocessing Queues are note reliable so switch to queue.queue instead inside unittest.
https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.empty


sebastianpfischer (approved):
Interesting :-/

Issue #21, opened 2022-02-18
Specify default var in yaml


sebastianpfischer:
As a *automated component test developer* I want to *be able to specify default values of ENV{} variables in the .yaml files*
so that *I can run tests locally without having to explicitly set the specific env variables as well as on a jenkins machine that uses the env variables.*

The syntax could for example be something similar to:
enable_key_exchange: ENV\{ENABLE_KEY_EXCHANGE=True}

*Acceptance Criteria*
AC1: Possible to add default values to env variables in the .yaml files
AC2: Some refactoring shall be discussed and done to reduce the cyclometric complexity.

PR #26, opened 2022-02-23
add the possibility to set default values to ENV variables in the yaml config files

TedRio:
Linked to: https://github.com/eclipse/kiso-testing/issues/21

Allow to set default values for env variables in case they are not found with the following syntax:
ENV{ENABLE_KEY_EXCHANGE=True}

### DIFF
--- a/src/pykiso/config_parser.py
+++ b/src/pykiso/config_parser.py
@@ -189,8 +189,8 @@ def parse_config(fname: PathType) -> typing.Dict:
                 env = match_env_with_val
             else:
                 raise ValueError(
-                        f"Environment variable {match_single_env} not found and no default value specified"
-                )   
+                    f"Environment variable {match_single_env} not found and no default value specified"
+                )
             is_numeric = re.fullmatch(r"\d+", env)
             is_hex = re.fullmatch(r"0x[0-9a-fA-F]+", env)
             is_bool = env.lower() in ["true", "false"]
--- a/src/pykiso/config_parser.py
+++ b/src/pykiso/config_parser.py
@@ -29,7 +29,6 @@ import typing
 from distutils.version import LooseVersion
 from pathlib import Path
 
-import elementpath
 import pkg_resources
 import yaml
 
--- a/src/pykiso/config_parser.py
+++ b/src/pykiso/config_parser.py
@@ -28,6 +28,7 @@ import sys
 import typing
 from distutils.version import LooseVersion
 from pathlib import Path
+import elementpath
 
 import pkg_resources
 import yaml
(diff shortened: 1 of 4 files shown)


src/pykiso/config_parser.py
```diff
@@ -162,20 +162,35 @@ def _fix_types_loc(thing: dict) -> dict:
...
         :return: config sub-dict value with replaced environment variable if needed
         """
+        has_default_value = False
+        # Handle default value
+        if all(element in config_element for element in ["ENV", "="]):
+            default_value = re.search("=(.*)}", config_element).group(1)
+            config_element = config_element[: config_element.find("=")] + "}"
+            has_default_value = True
         # Check for regular expression "ENV{word}"
         match = re.compile(r"ENV{(\w+)}").findall(str(config_element))
```
sebastianpfischer, line 181:
```suggestion
        match_single_env = re.compile(r"ENV{(\w+)}").findall(str(config_element))
        match_env_with_val = re.compile(r"ENV{(\w+)=(.+)}").findall(str(var2)) # search for the pattern
        
        
        
```

src/pykiso/config_parser.py
```diff
@@ -162,20 +162,35 @@ def _fix_types_loc(thing: dict) -> dict:
...
         if match:
```
sebastianpfischer, line 182: I'm surprised that match contains something and that you go in

sebastianpfischer (changes requested):

src/pykiso/config_parser.py
sebastianpfischer: I even thing it may be feasible to group them in one commando with '|' but I'm not good enough to find the right regex

Pog3k: you mean  
(\w+)|=(.+) 
?? ;D
The check if matching group 2 exist...

Pog3k: or maybe just this
ENV{(\w+)(=(.+))?}

sebastianpfischer: Have you tried them? (the second is better) I struggled on my side to match correctly the 2 possibilities

src/pykiso/config_parser.py
TedRio: It does, but it is much simpler to do as you said above

src/pykiso/config_parser.py
TedRio: Modified using ENV{(\w+)(=(.+))?}

src/pykiso/config_parser.py
```diff
@@ -162,20 +162,35 @@ def _fix_types_loc(thing: dict) -> dict:
...
 
         Additionally, cast the value if it matches an integer.
 
         :param config_element: config sub-dict value
 
         :return: config sub-dict value with replaced environment variable if needed
         """
+        has_default_value = False
+        # Handle default value
+        if all(element in config_element for element in ["ENV", "="]):
```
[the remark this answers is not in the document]
TedRio: Modified using only  the match

src/pykiso/config_parser.py
```diff
@@ -162,20 +163,35 @@ def _fix_types_loc(thing: dict) -> dict:
...
-        match = re.compile(r"ENV{(\w+)}").findall(str(config_element))
+        # Check for regular expression "ENV{word=.}"
+        match = re.compile(r"ENV{(\w+)(=(.+))?}").findall(str(config_element))
         if match:
             # Parse detected environment variable
-            match = str(match[0])
-            env = os.environ[match]
+            match_single_env = str(match[0][0])
+            match_env_with_val = str(match[0][2])
+            try:
```
sebastianpfischer, line 185:
```suggestion
            if match_single_env in os.environ;
                env = os.environ[match_single_env]
            elif match_env_with_val:
                env = match_env_with_val
            else:
                raise ValueError(
                        f"Environment variable {match_single_env} not found and no default value specified"
```

sebastianpfischer (changes requested):

src/pykiso/config_parser.py
TedRio: Done

sebastianpfischer (approved):

PR #29, opened 2022-03-01
feat: add uptime for pcan inside python-can

Pog3k:
will add uptime library for pcan timestamp

* changelog tool currently broke so ignore Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ invoke changelog
 
 #### Fixes
 
-* wait for logger thread to quit gracefully before closing segger channel
+* wait for logger thread to quit gracefully before closing segger channel ([#28](https://github.com/eclipse/kiso-testing/issues/28))
 #### Docs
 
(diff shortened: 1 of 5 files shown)


sebastianpfischer (changes requested):
I'm not sure, can you explain a little bit more what happened upstream that we need to add pcan this way? Before it was in the may library no?

sebclrsn (approved):

Pog3k:
> I'm not sure, can you explain a little bit more what happened upstream that we need to add pcan this way? Before it was in the may library no?

Actually its an ugly solution in python-can. I think it was never in python-can by default.
In the older version the just didn't print the import error.

See:
https://github.com/hardbyte/python-can/blob/develop/setup.py 
    "pcan": ["uptime~=3.0.1"], 
https://github.com/hardbyte/python-can/pull/1093

kakooloukia (approved):

sebastianpfischer (approved):

PR #34, opened 2022-03-22
feat: update to new release 0.16.0

Pog3k:


CHANGELOG.md
```diff
@@ -19,12 +25,14 @@ invoke changelog
 * Change auto changelog tool ([#30](https://github.com/eclipse/kiso-testing/issues/30))
 #### Continous integration
 
+* add skip lock for pipenv to prevent hangup
 * github-action for codecov ([#27](https://github.com/eclipse/kiso-testing/issues/27))
 #### Others
 
+* add skip lock for pipenv to prevent hangup
 * github-action for codecov ([#27](https://github.com/eclipse/kiso-testing/issues/27))
```
sebastianpfischer, line 33: This feels like a duplicate from above ^^

sebastianpfischer (changes requested):

CHANGELOG.md
Pog3k: yup, 
The file is managed entirely by auto-changelog.

Makes no sense to cleanup this file manualy.
Maybe this stuff will be fixt in future releases.

BKaDamien (approved):

CHANGELOG.md
sebastianpfischer: Can we not find out why?

sebclrsn (approved):

CHANGELOG.md
Pog3k: Suggestion would be to do it in a different task.
Its the python auto-changelog tool which then needs to be debugged.

The file will always be complete overwritten so we don't have to fix it in this task. 

IMO the release is more important.

sebastianpfischer (approved):

PR #42, opened 2022-04-07
feat: Make proxy agnoastic of transferred messages

sebastianpfischer:


src/pykiso/lib/auxiliaries/proxy_auxiliary.py
```diff
@@ -227,31 +227,25 @@ def _run_command(self) -> None:
...
-                self.channel._cc_send(*args, **kwargs)
-
-    def _dispatch_command(self, message: bytes, con_use: CChannel, remote_id: int):
+                any_msg, raw = conn.queue_in.get()
+                if any_msg is not None:
+                    self._dispatch_command(message=any_msg, con_use=conn, raw=raw)
+
+                self.channel._cc_send(any_msg, raw)
+
+    def _dispatch_command(self, message: bytes, con_use: CChannel, raw: bool):
```
sebastianpfischer, line 236: https://github.com/eclipse/kiso-testing/pull/39/files#diff-16f3c3870bf853285ede3cd36de77c70de7303cfa0d223be9fe9b118270ef43aR235 missing here

src/pykiso/lib/connectors/cc_proxy.py
```diff
@@ -60,28 +60,28 @@ def _cc_close(self) -> None:
         self.queue_out = Queue()
         log.debug("Close proxy channel")
 
-    def _cc_send(self, *args: tuple, **kwargs: dict) -> None:
+    def _cc_send(self, any_msg: Any, raw: bool = False) -> None:
```
sebastianpfischer, line 63:
```suggestion
    def _cc_send(self, msg: Any, raw: bool = False) -> None:
```
--> Above layer use sometimes named parameter

src/pykiso/lib/auxiliaries/proxy_auxiliary.py
```diff
@@ -227,31 +227,25 @@ def _run_command(self) -> None:
...
-                    self._dispatch_command(
-                        message=message,
-                        remote_id=kwargs.get("remote_id"),
-                        con_use=conn,
-                    )
-
-                self.channel._cc_send(*args, **kwargs)
-
-    def _dispatch_command(self, message: bytes, con_use: CChannel, remote_id: int):
+                any_msg, raw = conn.queue_in.get()
```
BKaDamien, line 230: If you go for any kind of message you will have to handle the different type in the auxiliary. In addition why keeping the raw, just handle the TestApp decoding in the corresponding auxiliary.

src/pykiso/lib/auxiliaries/proxy_auxiliary.py
```diff
@@ -227,31 +227,25 @@ def _run_command(self) -> None:
...
         :param message: message to send
         :param con_use: current proxy connector where the command come from
-        :param remote_id: if CAN is used, CAN frame ID on which
-            the message will be sent
+        :param raw: if the message is raw or not (makes no sense)
         """
         for conn in self.proxy_channels:
             if conn != con_use:
-                conn.queue_out.put([message, remote_id])
+                conn.queue_out.put((message, raw))
```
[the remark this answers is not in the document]
BKaDamien: Not always , for CAN message given by python-can it's a CAN Message object  where you will find more than just the payload

src/pykiso/lib/connectors/cc_proxy.py
BKaDamien: This will fail when using CAN and passing remote id

src/pykiso/lib/connectors/cc_proxy.py
```diff
@@ -60,28 +60,28 @@ def _cc_close(self) -> None:
...
 
         :return: raw bytes and source when it exist. if queue timeout
             is reached return None
         """
 
         try:
-            raw_msg, source = self.queue_out.get(True, self.timeout)
-            log.debug(f"received at proxy level : {raw_msg} || source {source}")
-            return raw_msg, source
+            any_msg, raw = self.queue_out.get(True, self.timeout)
```
BKaDamien, line 83: You will have performance issues if you want to put several auxes to the proxy ones

src/pykiso/lib/auxiliaries/proxy_auxiliary.py
sebastianpfischer: That will probably be one of the next steps to do.

src/pykiso/lib/connectors/cc_proxy.py
sebastianpfischer: yes, we will need to change the other auxiliaries accordingly. This is still ok because the user will not see it. it is an internal breaking change

src/pykiso/lib/auxiliaries/proxy_auxiliary.py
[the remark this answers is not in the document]
sebastianpfischer: Agree, I got lost in what needs to be returned :)

src/pykiso/lib/auxiliaries/proxy_auxiliary.py
Pog3k: yep, raw needs to be gone..

Pog3k: yep, raw needs to be gone..

src/pykiso/lib/connectors/cc_proxy.py
```diff
@@ -60,28 +60,28 @@ def _cc_close(self) -> None:
         self.queue_out = Queue()
         log.debug("Close proxy channel")
 
-    def _cc_send(self, *args: tuple, **kwargs: dict) -> None:
+    def _cc_send(self, any_msg: Any, raw: bool = False) -> None:
         """Populate the queue in of the proxy connector.
 
         :param args: tuple containing positionnal arguments
```
Pog3k, line 66: docstring needs to be updated

src/pykiso/lib/connectors/cc_proxy.py
```diff
@@ -60,28 +60,28 @@ def _cc_close(self) -> None:
...
         :param kwargs: dictionary containing named arguments
         """
-        log.debug(f"put at proxy level: {args} {kwargs}")
-        self.queue_in.put((args, kwargs))
+        log.debug(f"put at proxy level: {any_msg} {raw}")
+        self.queue_in.put((any_msg, raw))
```
Pog3k, line 70: I think we should pass args, kwargs in every queue. Then we would be fully data agnostic.
raw or remote_id can then be part of the args or kwars.

sebastianpfischer:
PR deleted, it served its purpose to explain the concept

PR #44, opened 2022-04-13
Migrate to poetry

sebclrsn:
- Drop python 3.6 support (black requires >=3.6.2 and python-can does not have a pcan extra before 4.0.0, 4.0.0 being compatible with python 3.7+)
- remove setup.cfg, setup.py, Pipfile and .coveragerc and add pyproject.toml with all elements
- restore black 21.12b1 due to click version issues (black>=22 requires click>=8 but auto-changelog requires click>=7, <8)
- update docs


docs/getting_started/getting_started.rst
```diff
@@ -19,15 +19,15 @@ Install
...
+`Poetry <https://python-poetry.org/>`__ is more appropriate for
 developers as it automatically creates virtual environments.
 
 .. code:: bash
 
    git clone https://dev-bosch.com/bitbucket/scm/pea/integration-test-framework.git
    cd integration-test-framework
-   pipenv install --dev
-   pipenv shell
+   poetry install
```
BKaDamien, line 29: we are in thee user installation s maybe the dev dependencies are not needed and user has to write : 
poetry install --no-dev

README.md
```diff
@@ -17,21 +17,20 @@ The project will contain:
...
 ```
 
-[Pipenv](https://github.com/pypa/pipenv) is more appropriate for developers as it automatically creates virtual environments.
+[Poetry](https://python-poetry.org/) is more appropriate for developers as it automatically creates virtual environments.
 
 ```bash
 cd kiso-testing
-pipenv install --dev
-pipenv shell
+poetry install
```
BKaDamien, line 32: we are in thee user installation s maybe the dev dependencies are not needed and user has to write :
poetry install --no-dev

.pre-commit-config.yaml
```diff
@@ -29,6 +29,7 @@ repos:
           - "--max-complexity=18"
           - "--select=B,C,E,F,W,T4,B9"
           - "--ignore=E203,E266,E501,W503,F401,E402"
+          - "--exclude=docs"
```
BKaDamien, line 32: do you really need this lines if you specify files: ^src/pykiso

sebclrsn: It was in the setup.cfg, so I just migrated the contents 1 to 1

BKaDamien: strange the setup.cfg was only there for the pytest configuration

README.md
sebclrsn: I agree for the getting_started, but here we state "is more appropriate for developers", so from what I understan this part would be for developers/contributors

Additionnally, we previously had pipenv install --dev here

docs/getting_started/getting_started.rst
BKaDamien: So then maybe delete this part from the getting_started because we have a page dedicated to devs.

sebclrsn: Even though I agree with you, it's still weird to skip the dev dependencies as right after we show the tasks to invoke, and invocations is part of the dev dependencies...

sebclrsn: But you made me notice that we still had old bitbucket links in there, thanks! :D

sebclrsn: I replaced the entire dev part with a simple pykiso example with options

BKaDamien (approved):

kakooloukia (approved):
@sebclrsn, Can you confirm that if master progress, it is still possible to point to the latest version ?
e.g. :
Considering:
package_A is using poetry
package_A depend on kiso-testing@LASTEST_HASH

sebclrsn:
> @sebclrsn, Can you confirm that if master progress, it is still possible to point to the latest version ? e.g. : Considering: package_A is using poetry package_A depend on kiso-testing@LASTEST_HASH

If you mean providing the git link instead of a fixed version, this is of course possible!

BKaDamien (approved):

sebastianpfischer (approved):

PR #51, opened 2022-05-06
feat: Make bus error warnings switchable for pcan

TedRio:


src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -34,6 +34,11 @@
 log = logging.getLogger(__name__)
 
 
+class PcanFilter(logging.Filter):
+    def filter(self, record):
```
kakooloukia, line 38: typehint + docstring missing

kakooloukia (changes requested):

src/pykiso/lib/connectors/cc_pcan_can.py
TedRio: Typehint and docstring have been added

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -34,6 +34,20 @@
 log = logging.getLogger(__name__)
 
 
+class PcanFilter(logging.Filter):
+    """Filter specific pcan logging messages"""
+
+    def filter(self, record: logging.LogRecord):
```
kakooloukia, line 40:
```suggestion
    def filter(self, record: logging.LogRecord) -> bool:
```

kakooloukia (changes requested):

kakooloukia (approved):

PR #66, opened 2022-05-20
feat: Rework pcan trace configuration in connector

Mcthomas777:
Added : 
- ability to change size of the PCAN trace
- ability to select a path where the trace must be
- ability to choose segmented or single trace (default : segmented)


src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -143,22 +152,15 @@ def __init__(
         generic logfile in the current working directory, which will be
         overwritten every time, a log is initiated.
         """
+        if self.trace_path.is_file():
+            self.trace_path = self.trace_path.parent
+            log.warning(
+                f"Input path is a file. Trace will be save into following path {self.trace_path}"
```
Pog3k, line 158: why warning ? Sounds that we don't suggest to use logfile name

sebclrsn: The reason is that the PCAN Basic API doesn't allow to specify a file name for the trace, hence the warning
Maybe be a bit more specific in your warning message e.g. `f"File names are not supported for trace file creation. Trace will be written to {self.trace_path}"`

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -80,6 +83,9 @@ def __init__(
         :param interface: python-can interface modules used
         :param channel: the can interface name
         :param state: BusState of the channel
+        :param trace_path: path to write the trace
+        :param trace_size: maximum size of the trace in (Mb)
+        :param segmented: Traced data is distributed in several files with size trace_size
```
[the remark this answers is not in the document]
Mcthomas777: I was a bit confused regarding this option, because I do not see any point to use single trace, if you can use segmented option with larger trace size. I just assumed that it was more flexible this way and as it does not request more ressource, may be it's better to let it like this ? 
However, I'l still add the check of trace size :)

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -80,6 +83,9 @@ def __init__(
         :param interface: python-can interface modules used
         :param channel: the can interface name
         :param state: BusState of the channel
+        :param trace_path: path to write the trace
+        :param trace_size: maximum size of the trace in (Mb)
```
sebclrsn, line 87:
```suggestion
        :param trace_size: maximum size of the trace (in MB)
```
Maybe also specify the possible values (1 to 100)

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -199,29 +207,52 @@ def _pcan_configure_trace(self) -> None:
         No exception is thrown in this case.
         """
         pcan_channel = getattr(PCANBasic, self.channel)
-        if self.logging_path is None:
+        if self.trace_path is None:
+            log.warning("No path specified, might overwrite existing trace.")
```
sebclrsn, line 211:
```suggestion
            log.warning("No trace path specified, an existing trace will be overwritten.")
```

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -199,29 +207,52 @@ def _pcan_configure_trace(self) -> None:
...
+            else:
+                log.info("Single option of trace file activated.")
+                self._pcan_set_value(
+                    pcan_channel,
+                    PCANBasic.TRACE_FILE_SINGLE,
+                    PCANBasic.PCAN_PARAMETER_ON,
+                )
+
+            if self.trace_size != 10:
+                log.info(f"Trace size set to: {self.trace_size}")
```
sebclrsn, line 244:
```suggestion
                log.info(f"Trace size set to {self.trace_size} MB")
```

sebclrsn (approved):

src/pykiso/lib/connectors/cc_pcan_can.py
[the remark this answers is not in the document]
Pog3k: Maybe I don't understand the option .

What happens If it don't set segmented and  trace_size is 1 MB but my Logfile has data for 10 MB.

Would it then just create one file with 1 MB and I loose the rest of the log ?  If you keep it like this then mention that information may get loose.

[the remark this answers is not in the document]
Mcthomas777: It will just save the first MB of information then once limit has been reached, nothing will be saved anymore :)

PR #90, opened 2022-07-08
docs: align dependencies in NOTICE.md

TedRio:


NOTICE.md
```diff
@@ -23,6 +23,9 @@ Thanks goes to these wonderful people (sorted by last name):
  * Yoann Garcia
```
sebastianpfischer, line 23: We should remove the spaces at the beginning but also sort them by last name

sebastianpfischer (changes requested):

NOTICE.md
TedRio: Done

sebastianpfischer (approved):

PR #92, opened 2022-07-12
bugfix: adapt rtt connector to create log folder if does not exist

TedRio:


sebastianpfischer:
I think we still should write a test for it no?

sebastianpfischer:
@TedRio, it does, but I think we can trust pathlib to do the job ^^

sebastianpfischer (approved):

sebclrsn (approved):

PR #98, opened 2022-07-14
Features/step report cross domain unit test

stg2sf:


CHANGELOG.md
```diff
@@ -129,6 +129,9 @@ Bugfix:
 Changes:
 - refactor test_case, test_suite, test_message_handler module
 
+Features:
```
BKaDamien, line 132: you don't need to speciy it just respect the commit rules

BKaDamien: In addition you have to update the pyprojecttoml with the include

setup.py
```diff
@@ -0,0 +1,96 @@
+##########################################################################
```
BKaDamien, line 1: this file is outdated and don't even exist

BKaDamien (changes requested):

sebastianpfischer (changes requested):
So from my understanding, https://github.com/eclipse/kiso-testing/pull/101 does the same no?

sebastianpfischer:
I close the PR because it is covered by #101

PR #107, opened 2022-07-25
Feature/allow tp waiting time customization

BKaDamien:
Just expose the tpwaitingtime param form python-uds in order to customize it when long/heavy UDS data exchange is needed


BKaDamien:
I never touch the message.py test so I don't really understand why the coverage is - 1.32% on this module

src/pykiso/lib/auxiliaries/udsaux/uds_auxiliary.py
```diff
@@ -73,7 +76,7 @@ def send_uds_raw(
             log.info(
                 f"UDS request to send '{['0x{:02X}'.format(i) for i in msg_to_send]}'"
             )
-            resp = self.uds_config.send(msg_to_send)
+            resp = self.uds_config.send(msg_to_send, tpWaitTime=tp_waiting_time)
```
sebclrsn, line 79: Could the tp_waiting_time eventually be added as UdsBaseAuxiliary attribute so that it can be set for the entire test and applied it the UDSServerAuxiliary also? 
https://github.com/eclipse/kiso-testing/blob/master/src/pykiso/lib/auxiliaries/udsaux/uds_server_auxiliary.py#L142

BKaDamien: As far I remember, If you play the rol of UDS server you only send single frame no?
If yes you never reach the else condition (see encode iso-tp from kiso-testing-python-uds in CanTP module)

sebclrsn: Actually, if the response to send by the server is bigger than 64 bytes, it will send the first frame and enter the wait state and the 10ms sleep could occur if no flow control is received by the client

Pog3k (approved):

src/pykiso/lib/auxiliaries/udsaux/uds_server_auxiliary.py
```diff
@@ -139,7 +139,9 @@ def send_response(self, response_data: List[int]) -> None:
         :param response_data: the UDS response to send.
         """
         to_send = self.uds_config.tp.encode_isotp(
-            response_data, use_external_snd_rcv_functions=True
+            response_data,
```
sebclrsn, line 142: As we discussed when testing, could you also please fix my bug by adding the `callback` parameter in `register_callback` (so l.196 and l.220)?

BKaDamien: Done

sebclrsn (approved):

Pog3k (approved):
Please Squash before merge

sebastianpfischer (commented):
I lack the purpose of this PR sorry

sebastianpfischer (changes requested):
I do not see how something is exposed...
See resp = self.uds_config.send(msg_to_send, tpWaitTime=self.tp_waiting_time)
To add the capability to play with the tp waiting time I  just create a the attributes self.tp_waiting_time (both server uds and client UDS can use it even if this used for the server version).

sebastianpfischer (approved):

PR #111, opened 2022-08-02
feat: Improve error reporting in test case execution

Mcthomas777:


src/pykiso/test_coordinator/test_execution.py
```diff
@@ -252,6 +252,9 @@ def execute(
         log.exception("Keyboard Interrupt detected")
         exit_code = ExitCode.ONE_OR_MORE_TESTS_RAISED_UNEXPECTED_EXCEPTION
     except Exception:
-        log.exception(f'Issue detected in the test-suite: {config["test_suite_list"]}!')
+        log.exception(
```
sebclrsn, line 255: I don't get this, the principle of log.exception is to log the message at error level AND to print out the exception stack, so even without this change you'd see all of the errors

src/pykiso/exceptions.py
```diff
@@ -14,12 +14,14 @@
...
 .. currentmodule:: exceptions
 """
+import logging
 
 
 class PykisoError(Exception):
     """Pykiso specific exception used as basis for all others."""
 
     def __str__(self):
+        logging.exception(self.message)
```
sebclrsn, line 24: As discussed don't mix up logging and exceptions, just add the log.exception when the exceptions are caught

sebclrsn (changes requested):

src/pykiso/exceptions.py
Mcthomas777: Ok, right just wanted to display the full message, as I didn't know it will be anyway

src/pykiso/exceptions.py
```diff
@@ -14,6 +14,7 @@
 
 .. currentmodule:: exceptions
 """
+import logging
```
sebastianpfischer, line 17: Is this import really necessary?

sebastianpfischer (changes requested):
Besides the above question, it looks good :)

tests/test_config_registry_and_test_execution.py
```diff
@@ -61,8 +64,14 @@ def test_config_registry_and_test_execution_with_pattern(tmp_test, capsys):
...
 
 
-@pytest.mark.parametrize("tmp_test", [("aux1", "aux2", False)], indirect=True)
-def test_config_registry_and_test_execution_collect_error(tmp_test, capsys, mocker):
+@pytest.mark.parametrize(
+    "tmp_test",
+    [("collector_error_aux1", "collector_error_aux_2", False)],
+    indirect=True,
+)
+def test_config_registry_and_test_execution_collect_error(
```
sebclrsn, line 72:
```suggestion
def test_config_registry_test_collection_error(
```
Just to split this test from the one you added

tests/test_config_registry_and_test_execution.py
```diff
@@ -87,6 +96,58 @@ def test_config_registry_and_test_execution_collect_error(tmp_test, capsys, mock
     assert "Ran 0 tests" not in output.err
 
 
+@pytest.mark.parametrize(
+    "tmp_test",
+    [("collector_error_2_aux1", "collector_error_2_aux", False)],
+    indirect=True,
+)
+def test_config_registry_and_test_execution_collect_error_2(mocker, caplog, tmp_test):
```
sebclrsn, line 104:
```suggestion
def test_config_registry_and_test_execution_collect_error(mocker, caplog, tmp_test):
```

src/pykiso/exceptions.py
Mcthomas777: Not at all

sebclrsn (approved):

sebastianpfischer (approved):

PR #113, opened 2022-08-05
bugfix: do not wait for an UDS response in tester present sender

TedRio:


src/pykiso/lib/auxiliaries/udsaux/uds_auxiliary.py
```diff
@@ -73,13 +75,16 @@ def send_uds_raw(
             log.info(
                 f"UDS request to send '{['0x{:02X}'.format(i) for i in msg_to_send]}'"
             )
-            resp = self.uds_config.send(msg_to_send)
+            resp = self.uds_config.send(msg_to_send, responseRequired=response_required)
```
sebastianpfischer, line 78: https://github.com/richClubb/python-uds/blob/master/uds/uds_communications/Uds/Uds.py#L142 kind of shocked me! So how many ticks do we wait for it to return?

tests/test_uds_auxiliary.py
```diff
@@ -327,5 +335,6 @@ def test_tester_present_sender(self, uds_raw_aux_inst, mocker):
             sleep(3)
 
         send_mock.assert_called_with(
-            UDSCommands.TesterPresent.TESTER_PRESENT_NO_RESPONSE
+            UDSCommands.TesterPresent.TESTER_PRESENT_NO_RESPONSE,
+            response_required=False,
```
sebastianpfischer, line 339: What about a TC where it is True?

sebastianpfischer (changes requested):

BKaDamien (approved):

tests/test_uds_auxiliary.py
TedRio: It is tested in other unit tests, but for the tester present sender it will always be False, as we do not expect an answer

sebastianpfischer: ok

sebclrsn (approved):

src/pykiso/lib/auxiliaries/udsaux/uds_auxiliary.py
TedRio: In this case we don't wait, if a response is required, the default value is either 1s, which seems a lot, or 1ms (the comment and the variable name are contradictory) : https://github.com/eclipse/kiso-testing-python-uds/blob/master/uds/uds_communications/TransportProtocols/Can/CanTp.py#L251

sebastianpfischer (approved):

PR #124, opened 2022-08-22
feat: add cc_serial connector

Pog3k:


src/pykiso/lib/connectors/cc_serial.py
```diff
@@ -0,0 +1,181 @@
...
+        """Read bytes from the serial port.
+        Try to read one byte in blocking mode. After blocking read check
+        remaining bytes and read them without a blocking call.
+
+        :param timeout: timeout in seconds, defaults to 0.00001
+        :param raw: not implemented, defaults to False
+        :raises NotImplementedError: if raw is to True
+        :return: received bytes
+        """
+        if raw:
```
sebastianpfischer, line 148: This is a mistake, raw = True means that bytes are transferred and not a message.

src/pykiso/lib/connectors/cc_serial.py
```diff
@@ -0,0 +1,181 @@
...
+        self.serial.timeout = 0
+
+        # Check how many bytes are still left and read them at once
+        in_waiting = self.serial.in_waiting
+        if in_waiting > 0:
+            received += self.serial.read(in_waiting)
+
+        return received
+
+    def cc_readline(self, timeout=0):
```
sebastianpfischer, line 165: Why are this APIs needed?

sebastianpfischer (changes requested):

src/pykiso/lib/connectors/cc_serial.py
Pog3k: I use it for the Automationboard which makes communication very easy and handy.
Its  also a standard method in pyserial and I think that many other devices also ends their messages whith a newline character

sebastianpfischer: This will not work with the proxy, please comply here with the dictionary approach to forward the parameters to the read method directly and handle it there :)

src/pykiso/lib/connectors/cc_serial.py
Pog3k: raw true is now default

src/pykiso/lib/connectors/cc_serial.py
```diff
@@ -0,0 +1,165 @@
...
+
+        received = self.serial.read()
+
+        # We have waited already so set the timeout to non blocking
+        self.serial.timeout = 0
+
+        # Check how many bytes are still left and read them at once
+        in_waiting = self.serial.in_waiting
+        if in_waiting > 0:
+            received += self.serial.read(in_waiting)
```
sebclrsn, line 163: Will this be blocking if you keep the original timeout? 
Otherwise you could keep the timeout value and change it only if it is different from the default, thus reducing the number of port reconfigurations due to the timeout change

something like:
```
    def _cc_receive(self, timeout=DEFAULT_TIMEOUT, raw: bool = True) -> Dict[str, bytes]:
        if timeout != DEFAULT_TIMEOUT:
            self.serial.timeout = timeout

        received = self.serial.read()

        # Check how many bytes are still left and read them at once
        in_waiting = self.serial.in_waiting
        if in_waiting > 0:
            received += self.serial.read(in_waiting)
```

sebclrsn (changes requested):

src/pykiso/lib/connectors/cc_serial.py
```diff
@@ -0,0 +1,165 @@
+##########################################################################
```
sebclrsn, line 1: Example yaml and documentation (at least what's new) are missing

sebclrsn, line 1: Regarding the discussion we had about the fix in pykiso-tags, the fix has to be added here: https://github.com/eclipse/kiso-testing/blob/master/src/pykiso/tool/show_tag.py#L287

And simply is: `output = Path(output)`

src/pykiso/lib/connectors/cc_serial.py
Pog3k: added dict approach... remove readline

src/pykiso/lib/connectors/cc_serial.py
Pog3k: added

Pog3k: added docs and loopback example

src/pykiso/lib/connectors/cc_serial.py
Pog3k: added an self attribute to store the timeout. If user requests now a different timeout the port will be reconfigured.
Added self read and write timeout

src/pykiso/lib/connectors/cc_serial.py
```diff
@@ -0,0 +1,165 @@
...
+        rtscts: bool = False,
+        write_timeout: float = None,
+        dsrdtr: bool = False,
+        inter_byte_timeout: float = None,
+        exclusive: bool = None,
+        **kwargs
+    ):
+        """Init Serial settings
+
+        :param port: Device name
```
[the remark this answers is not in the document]
Pog3k: added

src/pykiso/lib/connectors/cc_serial.py
```diff
@@ -0,0 +1,165 @@
...
+        :param parity: Enable parity checking. Possible values:
+          PARITY_NONE, PARITY_EVEN, PARITY_ODD PARITY_MARK, PARITY_SPACE,
+          defaults to PARITY_NONE
+        :param stopbits:  Number of stop bits. Possible values:
+          STOPBITS_ONE, STOPBITS_ONE_POINT_FIVE, STOPBITS_TWO,
+          defaults to STOPBITS_ONE
+        :param timeout: Set a read timeout value in seconds, defaults to None
+        :param xonxoff: Enable software flow contro, defaults to False
+        :param rtscts: Enable hardware (RTS/CTS) flow control, defaults to False
+        :param write_timeout: _description_, defaults to None
```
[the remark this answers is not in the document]
Pog3k: added

src/pykiso/lib/connectors/cc_serial.py
```diff
@@ -0,0 +1,165 @@
...
+    STOPBITS_ONE_POINT_FIVE = serial.STOPBITS_ONE_POINT_FIVE
+    STOPBITS_TWO = serial.STOPBITS_TWO
+
+
+class CCSerial(connector.CChannel):
+    """Connector for serial devices"""
+
+    def __init__(
+        self,
+        port,
```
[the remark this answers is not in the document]
Pog3k: added

src/pykiso/lib/connectors/cc_serial.py
```diff
@@ -0,0 +1,165 @@
...
+        parity: Parity = Parity.PARITY_NONE,
+        stopbits: StopBits = StopBits.STOPBITS_ONE,
+        timeout: float = None,
+        xonxoff: bool = False,
+        rtscts: bool = False,
+        write_timeout: float = None,
+        dsrdtr: bool = False,
+        inter_byte_timeout: float = None,
+        exclusive: bool = None,
+        **kwargs
```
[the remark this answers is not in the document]
Pog3k: Not all (: but the optionals are now optionals ..

docs/whats_new/version_ongoing.rst
```diff
@@ -1,2 +1,6 @@
 Version ongoing
 ---------------
+
+New serial connector
+^^^^^^^^^^^^^^^^^^^^
+Added new connector for serial communication.
```
sebastianpfischer, line 6: Maybe add the name of ? like that we do not need to search everywhere the APIs of it?

examples/test_serial/test_serial.py
```diff
@@ -0,0 +1,60 @@
...
+    """In this test we try to receive the same message which we have send.
+    RX is connected to TX on a serial dongle.
+    """
+
+    def test_run(self):
+        """Executing serial loopback example, let's try to send
+        a message and receive it.
+        """
+        logging.info(
+            f"--------------- RUN: {self.test_suite_id}, {self.test_case_id} ---------------"
```
sebastianpfischer, line 45: That is not needed right?

src/pykiso/lib/connectors/cc_serial.py
```diff
@@ -0,0 +1,174 @@
...
+        :param msg: data to send
+        :param raw: unused
+        :param timeout: write timeout in seconds. None sets it to blocking,
+          defaults to None
+        :raises: SerialTimeoutException - In case a write timeout is configured
+          for the port and the time is exceeded.
+        """
+        if timeout != self.current_write_timeout:
+            self.current_write_timeout = timeout
+            self.serial.write_timeout = timeout
```
sebastianpfischer, line 141:
```suggestion
        self.current_write_timeout = timeout or self.current_write_timeout
        self.serial.write_timeout = self.current_write_timeout
```
?

sebastianpfischer (changes requested):

src/pykiso/lib/connectors/cc_serial.py
Pog3k: The idea was to call self.serial.write_timeout only when necessary. -> reduce serial port reconfiguration
In your suggestion i would call the setter every time.

examples/test_serial/test_serial.py
Pog3k: removed.. added info messages which suites more to the serial test

sebclrsn (approved):

sebastianpfischer (approved):

PR #165, opened 2022-09-22
feat: process connector

cody82:


sebastianpfischer (changes requested):

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        try:
+            while True:
+                if self.text:
+                    data = stream.readline()
+                else:
+                    data = stream.read(1)
+                if len(data) == 0:
+                    break
+                # print(data)
```
Pog3k, line 161: comment ?

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def _cc_send(self, msg: MessageType, raw: bool = False, **kwargs) -> None:
+        """Execute process commands or write data to stdin"""
+        if isinstance(msg, dict) and "command" in msg:
+            if msg["command"] == "start":
+                self.start(msg["executable"], msg["args"])
+        elif self.pipe_stdin:
+            if self.process is None:
+                raise CCProcessError("Process is not running.")
+            log.debug(f"write stdin: {msg}")
```
Pog3k, line 181: log internal ?

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+        messages = messages[:i]
+        if len(messages) == 0:
+            return []
+
+        # Join messages
+        r = [ProcessMessage(messages[0].stream, b"".join([x.data for x in messages]))]
+        return r
+
+    def _cc_receive(self, timeout: float = 0.0001, raw: bool = False) -> MessageType:
+        """Receive messages"""
```
Pog3k, line 238: args docs missing

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+        :param stream: The stream to read from
+        :param name: The name of the stream
+
+        :return: The thread object
+        """
+        thread = threading.Thread(
+            name=f"cc_process_{name}", target=self._read_thread, args=(stream, name)
+        )
+        thread.start()
+        return thread
```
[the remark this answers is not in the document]
cody82: Performance is of course worse than reading directly from stdout/stderr, but as _cc_receive does not have a parameter to select the stream, this seems to be the best solution to keep latency low without a lot of polling.

tests/test_cc_process.py
```diff
@@ -0,0 +1,130 @@
+##########################################################################
```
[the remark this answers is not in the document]
cody82: The threads get stuck if the process gets stuck. But _cc_close will terminate the process and the threads in that case.

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+        :return: Existing messages as tuple: (streamname, data)
+        """
+        messages = self.buffer
+
+        # Get all messages from the process that are available
+        while not self.queue_in.empty():
+            messages.append(self.queue_in.get_nowait())
+        i = 1
+        # Find messages from the same stream(first entry in the tuple) as the first message
+        while (
```
sebclrsn, line 232: Wouldn't it be enough to find the first occurrence of None in the list?
`last_msg_idx = messages.index(None)`

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+
+        try:
+            read = self.queue_in.get(True, timeout)
+        except queue.Empty:
+            # Get previously received messages when in binary mode
+            existing = [] if self.text else self._read_existing()
+            if len(existing) > 0:
+                r = {"msg": {existing[0].stream: existing[0].data}}
+            else:
+                r = {"msg": None}
```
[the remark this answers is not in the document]
cody82: Done

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+            self.queue_in = None
+
+    def _cc_open(self) -> None:
+        """Implement abstract method"""
+        pass
+
+    def _read_existing(self) -> List[ProcessMessage]:
+        """Read buffered messages that where already received from the process
+
+        :return: Existing messages as tuple: (streamname, data)
```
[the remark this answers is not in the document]
cody82: Done

src/pykiso/lib/connectors/cc_process.py
cody82: Set to log internal debug.

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+        # Save the remaining messages for next time
+        self.buffer = messages[i:]
+
+        # Process only messages from the same stream
+        messages = messages[:i]
+        if len(messages) == 0:
+            return []
+
+        # Join messages
+        r = [ProcessMessage(messages[0].stream, b"".join([x.data for x in messages]))]
```
sebclrsn: Why do you need to return a list if it has at most 1 item?

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+                self.ready += 1
+                if self.ready == int(self.pipe_stdout) + int(self.pipe_stderr):
+                    # None marks the termination of all read threads
+                    self.queue_in.put(None)
+
+    def _cc_close(self) -> None:
+        self._cleanup()
+
+    def _cc_send(self, msg: MessageType, raw: bool = False, **kwargs) -> None:
+        """Execute process commands or write data to stdin"""
```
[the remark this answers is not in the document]
cody82: Done

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+                # print(data)
+                self.queue_in.put(ProcessMessage(name, data))
+        finally:
+            with self.lock:
+                self.ready += 1
+                if self.ready == int(self.pipe_stdout) + int(self.pipe_stderr):
+                    # None marks the termination of all read threads
+                    self.queue_in.put(None)
+
+    def _cc_close(self) -> None:
```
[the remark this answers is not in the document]
cody82: Done

src/pykiso/lib/connectors/cc_process.py
cody82: Done

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+
+MessageType = Union[str, ByteString]
+
+
+class CCProcessError(BaseException):
+    ...
+
+
+@dataclass
+class ProcessMessage:
```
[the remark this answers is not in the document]
cody82: Done

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0.
+#
+# SPDX-License-Identifier: EPL-2.0
+##########################################################################
+
+"""
+Process Channel
+*************
```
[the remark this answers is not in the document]
cody82: Done

examples/process.yaml
```diff
@@ -0,0 +1,23 @@
+auxiliaries:
+  com_aux:
+    connectors:
+        com: process1
+    type: pykiso.lib.auxiliaries.communication_auxiliary:CommunicationAuxiliary
+connectors:
+  process1:
+    config:
```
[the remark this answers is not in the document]
cody82: Done

sebclrsn (changes requested):

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+            self.process.stdin.write(msg)
+            self.process.stdin.flush()
+        else:
+            raise CCProcessError("Can not send to stdin because pipe is not enabled.")
+
+    def _cleanup(self):
+        """Cleanup threads and process objects"""
+        if self.process is not None:
+            self.process.terminate()
+            self.process.wait()
```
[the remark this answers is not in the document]
cody82: According to the documentation it will send SIGTERM. Do you mean that it should rather use SIGKILL to be sure?

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+        # Buffer for messages that where read from the process but not yet returned by _cc_receive
+        self.buffer: List[Optional[ProcessMessage]] = []
+
+    def start(self, executable: Optional[str] = None, args: Optional[List[str]] = None):
+        """Start a process
+
+        :param executable: The executable path. Default to path specified in yaml if not given.
+        :param args: The process arguments. Default to arguments specified in yaml if not given.
+
+        :return: The thread object
```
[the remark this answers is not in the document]
cody82: What do you mean? It says "return thread" on the bottom.

src/pykiso/lib/connectors/cc_process.py
cody82: Right, previously it returned more than one, but now thats not possible anymore. Will change that.

src/pykiso/lib/connectors/cc_process.py
cody82: No, possible values are: stdout, stderr, None. So if the buffer contains stdout, stdout, stderr, None it should return the first two entries. But there might be a more elegant way of doing this.

src/pykiso/lib/connectors/cc_process.py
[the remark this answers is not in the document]
sebclrsn: _start_read_thread returns à thread, but not start, it simply assigns the thread to the attribute (or I really need glasses)

src/pykiso/lib/connectors/cc_process.py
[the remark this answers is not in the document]
sebclrsn: I think so, or better try to gracefully terminate the process within timeout s and if this fails send a sigkill

src/pykiso/lib/connectors/cc_process.py
sebclrsn: You could use
```python
initial_stream = messages[0]
itertools.dropwhile(lambda m: m is not None and m.stream == initial_stream, messages)
```

src/pykiso/lib/connectors/cc_process.py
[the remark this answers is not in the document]
sebastianpfischer: I worry about the global performance impact. How much CPU does it use?

tests/test_cc_process.py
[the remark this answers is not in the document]
sebastianpfischer: Do you have a test for that?

src/pykiso/lib/connectors/cc_process.py
cody82: Correction: It can return empty list, list with one message, list with None. When removing the list I could not return nothing.

src/pykiso/lib/connectors/cc_process.py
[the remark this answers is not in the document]
cody82: Done

src/pykiso/lib/connectors/cc_process.py
cody82: I need both parts though. Basically, I want to split the list at the calculated index.

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,265 @@
...
+                if self.ready == int(self.pipe_stdout) + int(self.pipe_stderr):
+                    # None marks the termination of all read threads
+                    self.queue_in.put(None)
+
+    def _cc_close(self) -> None:
+        self._cleanup()
+
+    def _cc_send(self, msg: MessageType, raw: bool = False, **kwargs) -> None:
+        """Execute process commands or write data to stdin"""
+        if isinstance(msg, dict) and "command" in msg:
```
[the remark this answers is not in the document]
cody82: Done

src/pykiso/lib/connectors/cc_process.py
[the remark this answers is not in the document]
cody82: Fixed documentation

tests/test_cc_process.py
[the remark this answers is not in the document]
cody82: Test added.

Pog3k (approved):

src/pykiso/lib/connectors/cc_process.py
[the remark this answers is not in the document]
cody82: After discussion: Will add a kill if the process does not exit after a few seconds.

src/pykiso/lib/connectors/cc_process.py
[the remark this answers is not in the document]
cody82: Reads are blocking so it does not waste CPU.

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,319 @@
...
+    # Data as bytes or string
+    data: Union[str, bytes]
+
+
+class ProcessExit:
+    ...
+
+
+# Marker for process exit in queue_in
+PROCESS_EXIT = ProcessExit()
```
sebclrsn, line 59: Maybe make this private 
```suggestion
_PROCESS_EXIT_SENTINEL = ProcessExit()
```
Also, why making an additional ProcessExit class when you can just do this:
```python
_PROCESS_EXIT_SENTINEL = object()
```

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,319 @@
...
+        self.text = text
+        self.cwd = cwd
+        self.env = env
+        self.process: Optional[subprocess.Popen] = None
+        self.queue_in: Optional[queue.Queue[Union[ProcessMessage, ProcessExit]]] = None
+        self.stdout_thread: Optional[threading.Thread] = None
+        self.stderr_thread: Optional[threading.Thread] = None
+        self.lock: threading.Lock = threading.Lock()
+        # Counter for finished threads
+        self.ready: int = 0
```
sebclrsn, line 110: You don't need to put typehints everywhere, here you already assign an integer to it so the typehint doesn't bring any added value

sebclrsn, line 110: Also, if you'd use a more explicit name for the variable you wouldn't need to add a comment to explain its goal 

And IMO this should be private

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,319 @@
...
+        while (
+            i < len(messages)
+            and messages[0] is not PROCESS_EXIT
+            and messages[i] is not PROCESS_EXIT
+            and messages[0].stream == messages[i].stream
+        ):
+            i += 1
+
+        # Save the remaining messages for next time
+        self.buffer = messages[i:]
```
sebclrsn, line 256:
```suggestion
        messages, self.buffer = messages[:i], messages[i:]
```

src/pykiso/lib/connectors/cc_process.py
```diff
@@ -0,0 +1,319 @@
...
+        if self.process is not None:
+            # Terminate the process if still running
+            self.process.terminate()
+            try:
+                self.process.wait(5)
+            except subprocess.TimeoutExpired:
+                log.internal_warning(
+                    f"Process {self.executable} could not be terminated"
+                )
+                self.process.kill()
```
sebclrsn, line 218: On windows, kill does exactly the same as terminate so I don't think this will be useful

src/pykiso/lib/connectors/cc_process.py
sebclrsn: Then:
```python
initial_stream = messages[0].stream
consecutive_messages = list(itertools.dropwhile(lambda m: m is not None and m.stream == initial_stream, messages))
self.buffer = messages[len(consecutive_messages):]
```

If you really want to keep it simple another more readable alternative IMO would be:
```python
if messages[0] is PROCESS_EXIT:
    self.buffer = messages[1:]
    return None
    
message_end_idx  = len(messages)   
initial_stream = messages[0].stream

for i, message in enumerate(messages):
    if message is PROCESS_EXIT or message.stream != initial_stream:
        message_end_idx = i
        break
messages, self.buffer = messages[:message_end_idx], messages[message_end_idx:]
```

src/pykiso/lib/connectors/cc_process.py
cody82: Because then the type of the queue items is Any, which I did not like. Or does Union[ProcessMessage, object] make sense? Because everything is object?

src/pykiso/lib/connectors/cc_process.py
cody82: Done

cody82: Done

src/pykiso/lib/connectors/cc_process.py
cody82: Will not change that as it also does not cause a problem.

src/pykiso/lib/connectors/cc_process.py
cody82: Done

sebclrsn (approved):

PR #186, opened 2022-10-17
Bugfix/fix step report

BKaDamien:
always create StepReportData instance even if --step-report flag is not use to ensure the compatibility with test case using step_report variable

### DIFF
--- a/src/pykiso/test_coordinator/test_execution.py
+++ b/src/pykiso/test_coordinator/test_execution.py
@@ -187,7 +187,9 @@ def failure_and_error_handling(result: unittest.TestResult) -> int:
     return exit_code
 
 
-def enable_step_report(all_tests_to_run: unittest.suite.TestSuite) -> None:
+def enable_step_report(
+    all_tests_to_run: unittest.suite.TestSuite, step_report: Path
+) -> None:
     """Decorate all assert method from Test-Case
 
         This will allow to save the assert inputs in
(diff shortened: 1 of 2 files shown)


src/pykiso/test_coordinator/test_execution.py
```diff
@@ -199,20 +201,19 @@ def enable_step_report(all_tests_to_run: unittest.suite.TestSuite) -> None:
...
-        # Decorate All assert method
-        assert_method_list = [
-            method for method in dir(tc) if method.startswith("assert")
-        ]
-        for method_name in assert_method_list:
-            # Get method from name
-            method = getattr(tc, method_name)
-            # Add decorator to the existing method
-            setattr(tc, method_name, assert_decorator(method))
+        tc.step_report = StepReportData()
```
sebastianpfischer, line 204: I forgot we got that here... I do not like it

sebastianpfischer (approved):

src/pykiso/test_coordinator/test_execution.py
sebclrsn: Only other possibility would be to add it by default in BasicTest, BasicTestSuiteSetup and BasicTestSuiteTearDown

sebclrsn (approved):

src/pykiso/test_coordinator/test_execution.py
BKaDamien: if we do that we have to redesign a little bit our framework

BKaDamien: sorry it makes two words

sebastianpfischer: Yea... I was thinking really to cluster it into the test_result section in the future.
and the part where we decorate the `asserts`, we make it a little bit nicer and independent of the step reporter.

Pog3k (approved):

PR #200, opened 2022-11-23
Feature/fix misc issues

sebastianpfischer:


TedRio (approved):

src/pykiso/auxiliary.py
```diff
@@ -57,6 +57,7 @@ def __init__(self) -> None:
         self.is_instance = False
         self.stop_event = None
         self._aux_copy = None
+        self.connector_required = True
```
BKaDamien, line 60: due we use this interface since the DT implementation?

BKaDamien (approved):

src/pykiso/auxiliary.py
Pog3k: its new since we wanted to have connectors optional

PR #204, opened 2022-12-05
bump pykiso-python-uds version to ^3.0.2

JoSglch:
new release of pykiso-python-uds includes fix for dynamic uds message lengths


sebastianpfischer:
This is strange that the .lock file is not updated. In addition, I think you did not follow https://kiso-testing.readthedocs.io/en/latest/getting_started/contribution_guide.html which means that the Eclipse Foundation will not allow the merge.

PR #208, opened 2022-12-13
fix: fix skip test cases of a test suite if a test suite setup fails

yannpoupon:
Fix skipping test cases if a test suite setup fails by looking if an error occurred in the test and then storing it.
I use `result.error_occured` because it is how a test is considered failed or not : https://github.com/eclipse/kiso-testing/blob/master/src/pykiso/test_result/text_result.py#L199

### DIFF
--- a/src/pykiso/test_coordinator/test_suite.py
+++ b/src/pykiso/test_coordinator/test_suite.py
@@ -378,16 +378,15 @@ class BasicTestSuite(unittest.TestSuite):
     ) -> None:
         """Check if the suite setup has failed and store failed suite id.
         Search in the global unittest result object, which save all the results
-        of the tests performed up to that point, for a BasicTestSuiteSetup tests
-        which has failed. If the suite setup has failed store the suite id.
+        of the tests performed up to that point, if the suite setup that runned
+        has failed then store the suite id.
 
         :param test: test to check
         :param result: unittest result object
         """
         if isinstance(test, BasicTestSuiteSetup):
-            for suite_type, _ in result.failures:
-                if isinstance(suite_type, BasicTestSuiteSetup):
-                    self.failed_suite_setups.add(test.test_suite_id)
+            if result.error_occurred:
+                self.failed_suite_setups.add(test.test_suite_id)
 
     def run(self, result: BannerTestResult, debug: bool = False) -> BannerTestResult:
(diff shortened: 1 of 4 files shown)


src/pykiso/test_coordinator/test_suite.py
```diff
@@ -378,16 +378,15 @@ def check_suite_setup_failed(
...
+        has failed then store the suite id.
 
         :param test: test to check
         :param result: unittest result object
         """
         if isinstance(test, BasicTestSuiteSetup):
-            for suite_type, _ in result.failures:
-                if isinstance(suite_type, BasicTestSuiteSetup):
-                    self.failed_suite_setups.add(test.test_suite_id)
+            if result.error_occurred:
```
Pog3k, line 388: Im wondering why the old code doesnt work anymore. did we update ab deps ?
anyways you approach also looks ok

Pog3k (approved):

src/pykiso/test_coordinator/test_suite.py
sebastianpfischer: What do you mean?

sebastianpfischer: But yes, the previous check was removed. Is there a reason for that?

sebastianpfischer (approved):

src/pykiso/test_coordinator/test_suite.py
Pog3k: Somehow Im sure that the old code has worked and i dont get it why its now broken Oo

yannpoupon: The previous check was removed because my new check add the same suite id than looking in result failures, because when a test is added in `result.failures` it also put `result.error_occurred` to true. : https://github.com/eclipse/kiso-testing/blob/master/src/pykiso/test_result/text_result.py#L211-L221
One advantage is that if we have multiple suite it will not skip the second one if the first suite setup fails.

sebclrsn (approved):

src/pykiso/test_coordinator/test_suite.py
yannpoupon: The old code seems to be working if the suite setup has an assertion error but not if another type of error raised.

PR #213, opened 2022-12-15
refactor: granularity in the installation

sebastianpfischer:


pyproject.toml
```diff
@@ -41,23 +41,55 @@ classifiers = [
...
+    "pyserial",
+]
+all = [
+    "pylink-square",
+    "pykiso-python-uds",
+    "python-can",
+    "pyserial",
+    "PyVISA",
+    "PyVISA-py",
+]
```
kakooloukia, line 89: Is it possible to re-use variables with poetry ?
eg:
all = can + debugger + instrument + serial

what is the difference between 'all' and 'plugins' ?

sebastianpfischer: Hey!
So @Pog3k answered your question :)
For the last point. Next step would be to separate the different tools from the core part. Which mean that there will be a difference between `plugins` and `all`.

src/pykiso/lib/connectors/cc_serial.py
```diff
@@ -23,7 +23,11 @@
 from enum import Enum, IntEnum
 from typing import ByteString, Dict, Optional, Union
 
-import serial
+try:
+    import serial
+except ImportError("Dependencies missing, please install 'pip install pykiso[serial]'"):
+    raise
```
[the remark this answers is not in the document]
sebastianpfischer: Fair ^^, I will apply as soon as I have some time

BKaDamien (approved):

kakooloukia (approved):

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -24,9 +24,13 @@
 from pathlib import Path
 from typing import Dict, Optional, Union
 
-import can
-import can.bus
-import can.interfaces.pcan.basic as PCANBasic
+try:
```
sebclrsn, line 27:
```
mock_python_can = MagicMock(side_effect=ImportError)
with patch.dict('sys.modules', can=mock_python_can):
    import cc_pcan_can
```

sebastianpfischer: I have a pretty dangerous hack!!!!

```python
def test_import(mocker):
    with pytest.raises(ImportError):
        sys.modules['can'] = None
        importlib.reload(cc_pcan_can)
    sys.modules['can'] = can
    importlib.reload(cc_pcan_can)
```

sebastianpfischer:
The coverage is not reflecting the reality anymore, the improvement can be found here: https://app.codecov.io/gh/eclipse/kiso-testing/commit/58f0e4808752e6d05e39c3a236c8482720820074

sebastianpfischer:
Somehow there is a mismatch between the black versions used in Jenkins and the one we use... I could commit because it passed black formatting ^^ but Jenkins complains it does not.

sebastianpfischer:
Coverage of `cc_usb.py` should be done with #255

PR #215, opened 2022-12-21
Feature/proxy attribute forwarding

sebclrsn:
This PR implements the cchannel API forwarding mechanism discussed as partial solution 2 for complex connectors.

It also fixes the activate_log functionality so that when loggers are activated by one auxiliary, they aren't deactivated by the following ones

### DIFF
--- a/docs/advanced_usage/advanced_config_file.rst
+++ b/docs/advanced_usage/advanced_config_file.rst
@@ -131,6 +131,12 @@ that will have exclusive access to the communication channel. A
 the created :py:class:`~pykiso.lib.auxiliaries.proxy_auxiliary.ProxyAuxiliary`. This allows to
 keep a minimalistic :py:class:`~pykiso.connector.CChannel` implementation.
 
+Access to attributes and methods of the defined communication channel that has been
+attached to the created :py:class:`~pykiso.lib.auxiliaries.proxy_auxiliary.ProxyAuxiliary`
+is still possible, but keep in mind that if one auxiliary modifies one the communication
+channel's attributes, every other auxiliary sharing this communication channel will be
+affected by this change.
+
 An illustration of the resulting internal setup can be found at :ref:`proxy_aux`.
 
 In other words, if you define the following YAML configuration file:
--- a/docs/whats_new/version_ongoing.rst
+++ b/docs/whats_new/version_ongoing.rst
@@ -9,7 +9,7 @@ It is no longer necessary to manually defined a ``ProxyAuxiliary`` with
 each auxiliary that has to share it, ``pykiso`` will do the rest for you.
 
 
-Remove reduncancy of activate_log configuration parameter
+Remove redundancy of activate_log configuration parameter
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 In order to activate external loggers, it was previously necessary to
--- a/docs/whats_new/version_ongoing.rst
+++ b/docs/whats_new/version_ongoing.rst
@@ -7,13 +7,3 @@ Internal creation of proxy auxiliaries
 It is no longer necessary to manually defined a ``ProxyAuxiliary`` with
 ``CCProxy``s yourself. If you simply pass the communication channel to
 each auxiliary that has to share it, ``pykiso`` will do the rest for you.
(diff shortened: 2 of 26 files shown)


docs/whats_new/version_ongoing.rst
```diff
@@ -1,2 +1,19 @@
...
+
+Internal creation of proxy auxiliaries
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is no longer necessary to manually defined a ``ProxyAuxiliary`` with
+``CCProxy``s yourself. If you simply pass the communication channel to
+each auxiliary that has to share it, ``pykiso`` will do the rest for you.
+
+
+Remove reduncancy of activate_log configuration parameter
```
Pog3k, line 12:
```suggestion
Remove redundancy of activate_log configuration parameter
```

docs/advanced_usage/advanced_config_file.rst
```diff
@@ -131,6 +131,12 @@ that will have exclusive access to the communication channel. A
 the created :py:class:`~pykiso.lib.auxiliaries.proxy_auxiliary.ProxyAuxiliary`. This allows to
 keep a minimalistic :py:class:`~pykiso.connector.CChannel` implementation.
 
+Access to attributes and methods of the defined communication channel that has been
+attached to the created :py:class:`~pykiso.lib.auxiliaries.proxy_auxiliary.ProxyAuxiliary`
+is still possible, but keep in mind that if one auxiliary modifies one the communication
+channel's attributes, every other auxiliary sharing this communication channel will be
+affected by this change.
```
Pog3k, line 138: I think we should illustrate that with a python snippet somewhere

Pog3k (approved):

docs/whats_new/version_ongoing.rst
sebclrsn: To me it is sufficient to have this behaviour explained with text and then illustrated with the schema linked below, adding a code snippet here will add a lot of content for a side-note the user shouldn't usually have to care about

docs/advanced_usage/advanced_config_file.rst
sebclrsn: To me it is sufficient to have this behaviour explained with text and then illustrated with the schema linked below, adding a code snippet here will add a lot of content for a side-note the user shouldn't usually have to care about

docs/whats_new/version_ongoing.rst
```diff
@@ -1,2 +1,19 @@
...
+
+Internal creation of proxy auxiliaries
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is no longer necessary to manually defined a ``ProxyAuxiliary`` with
+``CCProxy``s yourself. If you simply pass the communication channel to
+each auxiliary that has to share it, ``pykiso`` will do the rest for you.
+
+
+Remove redundancy of activate_log configuration parameter
```
sebastianpfischer, line 12: Where is that coming from?

src/pykiso/auxiliary.py
```diff
@@ -26,9 +26,9 @@
 import time
 from typing import Any
 
-from .logging_initializer import add_logging_level
-from .test_setup.config_registry import ConfigRegistry
-from .types import MsgType
+from pykiso.logging_initializer import add_internal_log_levels
```
sebastianpfischer, line 29: Please cluster the features in the following way: one feature <-> one PR. Else, there is a miss-match between your PR title, why it exists and its content.

sebastianpfischer (changes requested):

sebastianpfischer:
Please separate the PR in 2 with clear descriptions for each and the feature you want to add. Thanks!

src/pykiso/auxiliary.py
sebclrsn: That's not a feature but an extremely small refactoring at auxiliary level that just uses an API from a previous PR

I'll make a separate PR for the activate_log feature modification, but this one can definitely stay in here IMO

docs/whats_new/version_ongoing.rst
sebclrsn: Now part of another PR

BKaDamien (approved):

PR #221, opened 2023-01-09
feat: add step report folding

Pog3k:
Tests are now foldable items.
If tests have failed they are open by default.

Also removed whitespaces when html gets generated

### DIFF
--- a/src/pykiso/test_result/assert_step_report.py
+++ b/src/pykiso/test_result/assert_step_report.py
@@ -37,7 +37,7 @@ from collections import OrderedDict
 from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path
-from typing import Union
+from typing import List, Union
 from unittest.case import TestCase, _SubTest
 
 import jinja2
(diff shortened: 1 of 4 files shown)


src/pykiso/test_result/assert_step_report.py
```diff
@@ -355,6 +355,20 @@ def _parse_timestamp(timestamp: float) -> str:
     return dt.strftime("%d/%m/%y %H:%M:%S")
 
 
+def is_test_success(test: List) -> bool:
+    """Check if test was successful.
+
+    :param tests: test
+    :return: True if each step in a test was successful else False
+    """
+    return not any([not step["succeed"] for step in test])
```
sebclrsn, line 364:
```suggestion
    return all(step["succeed"] for step in test)
```
This seems equivalent and more readable

sebclrsn (changes requested):

sebclrsn (approved):

BKaDamien (approved):

PR #230, opened 2023-01-23
feat: make release 0.21.0

yannpoupon:


docs/whats_new/version_0_21_0.rst
```diff
@@ -0,0 +1,62 @@
+Version 0.21.0
```
Pog3k, line 1: We added the new feature from Sebastian so it should also appear here.

Pog3k (approved):

BKaDamien (approved):

PR #233, opened 2023-01-23
Bugfix/dash in testcase pattern

JoSglch:
To prevent silently skipping tests with incorrect module names, throw an exception if a test_filter_pattern matches a file with an invalid python module name.

### DIFF
--- a/src/pykiso/exceptions.py
+++ b/src/pykiso/exceptions.py
@@ -62,11 +62,13 @@ class ConnectorRequiredError(PykisoError):
         super().__init__(self.message)
 
 
-class InvalidPattern(PykisoError):
-    """Raised when a invalid python module name is used as test file pattern by
-    the cli or when parsing the yaml config
+class InvalidTestModuleName(PykisoError):
+    """Raised when a invalid python module name is covered by test file pattern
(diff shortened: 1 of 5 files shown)


src/pykiso/exceptions.py
```diff
@@ -60,3 +60,15 @@ def __init__(self, aux_name: str) -> None:
...
+
+
+class InvalidTestModuleName(PykisoError):
+    """Raised when a invalid python module name is covered by test file pattern
+    of either the cli or the yaml config
+    """
+
+    def __init__(self, name: str) -> None:
+        self.message = (
+            f"Test files need to be valid python module names but '{name}' is invalid"
```
Pog3k, line 72: Also tell what are invalid symbols or characters. Or what is allowed
If i look into the pattern r'[_a-z]\w*\.py$'   
only lowercase words are allowed and the "_" char

JoSglch: The pattern used by unittest also allows upper case (it is compiled with re.IGNORECASE) and numbers (\w). Added hint.

Pog3k (approved):

PR #244, opened 2023-02-07
tests: adding test for show_tag and pykiso to pytest

yannpoupon:


pyproject.toml
```diff
@@ -59,10 +59,10 @@ pyreadline3 = {version = "^3.4.1", python = "^3.5"}
 hidapi = "^0.12.0.post2"
 rich = {version = "^13.2.0", optional = true}
 requests = {version = "^2.28.2", optional = true}
-
+isort = ">=5.11.4"
+black = "22.10.0"
```
sebastianpfischer, line 63: Why?

sebastianpfischer (changes requested):

pyproject.toml
yannpoupon: The pipeline with tox were having import issue and so the tests were not runned.
https://ci.eclipse.org/kiso-testing/job/kiso-testing/job/PR-244/3/console

The solution I found were changing the dependencies since pykiso to pytest need these dependencies : 
https://github.com/eclipse/kiso-testing/blob/master/src/pykiso/tool/pykiso_to_pytest/cli.py#L17-L18

pyproject.toml
```diff
@@ -115,7 +115,7 @@ omit = [
     "*/examples/*",
     "*/setup.py",
     "*/tasks.py",
-    "*/src/pykiso/tool/*",
+    "*/src/pykiso/tool/testrail/*",
```
BKaDamien, line 118: Why only exclude testrail tool just before we had all package in tool folder?

BKaDamien (changes requested):

pyproject.toml
yannpoupon: For the testrail it was just because some files like the cli.py and console.py for testrail had no test so I was not sure if I should just skip these files or all the repository for the code coverage.

pyproject.toml
sebclrsn: If you have to add these deps, create an extra called e.g. `pytest`

sebclrsn (approved):

pyproject.toml
yannpoupon: Done

BKaDamien (approved):

PR #249, opened 2023-02-15
fix: step report now display nested assertions

BKaDamien:
Remove parent method filtering from assert decorator in assert_step_report


sebastianpfischer (approved):

src/pykiso/test_result/assert_step_report.py
```diff
@@ -271,53 +271,51 @@ def func_wrapper(*args, **kwargs):
...
-
-                # 2. Update report data
-                # 2.1 Ensure report ready for update
-                _prepare_report(test_case_inst, test_name)
-
-                # 2.2. Add new step
-                _add_step(
-                    test_class_name, test_name, message, var_name, expected, received
-                )
+            # get the decorated test fixture name (setUp, tearDown, ...)
```
sebclrsn, line 282: After discussion, the nested assertions within method calls should all be part of the current running testcase, i.e. when parent_method doesn't start with `test_` then the call stack should be inspected until a valid candidate is found

BKaDamien: Perfect ,(. I will apply the changes

sebclrsn (approved):

src/pykiso/test_result/assert_step_report.py
```diff
@@ -232,6 +233,35 @@ def _add_step(
     )
 
 
+def determine_parent_test_function(test_name: str) -> str:
```
sebastianpfischer, line 236: Can you provide some more information about this function? More towards time-complexity, how fast are we or how slow did we get?

sebastianpfischer (approved):
Same as for the other PR, is it possible to get a before % after screenshot?

src/pykiso/test_result/assert_step_report.py
BKaDamien: how fast are we or how slow did we get? In the case of nested assert we get slower (more code to execute). Otherwise (assert method not nested) nothing really significant.

Based on the function and the call of the stack a few ms (10 -20 ms) more if the assert method is nested.

Otherwise (assert method not nested) a just the if comparison so a few micro seconds.

PR #252, opened 2023-02-21
step report shows each test method description

BKaDamien:


sebastianpfischer (approved):
Can you provide a screenshot of the output? like "before" and "after"?

PR #263, opened 2023-03-23
feat: handle logging for multiple yamls

TedRio:


src/pykiso/cli.py
```diff
@@ -94,6 +95,18 @@ def check_file_extension(
...
 
 
+def get_file_name_without_extension(path: PathType) -> str:
+    """Get name of a file without its extension
+
+    :param path: path of the file
+
+    :return: file name without extension
+    """
+    full_name = os.path.basename(path)
```
sebclrsn, line 105: With pathlib this would become:
```return Path(path).resolve().name```

File: src/pykiso/cli.py [hunk 2]
```diff
@@ -112,10 +125,11 @@ def check_file_extension(
 )
 @click.option(
     "-l",
-    "--log-path",
+    "--log-paths",
```
sebclrsn, line 128: I wouldn't introduce a breaking change on the name just because we are now supporting multiple ones, only the option's help should be extended IMO

src/pykiso/cli.py
```diff
@@ -204,9 +218,31 @@ def main(
...
+                    log_paths[0], log_level, verbose, report_type, yaml_name
+                )
+            # Log in different files for each yaml
+            elif len(log_paths) == len(test_configuration_file):
+                logger = initialize_logging(
+                    log_paths[log_index], log_level, verbose, report_type, yaml_name
+                )
+                log_index += 1
+            else:
+                raise ValueError(
```
sebclrsn, line 239: Consider using `click.UsageError` or any click CLI-specific exception instead

src/pykiso/logging_initializer.py
```diff
@@ -114,8 +116,10 @@ def filter(self, record):
     # if log_path is given create a file handler
     if log_path is not None:
         log_path = Path(log_path)
-        if log_path.is_dir():
-            fname = time.strftime("%Y-%m-%d_%H-%M-test.log")
+        if log_path.suffix == "":
+            if not log_path.is_dir():
```
sebclrsn, line 120: Maybe more 
```
if log_path.is_dir():
    log_path.mkdir(exist_ok=True, parent=True)
``` 
?

sebclrsn (changes requested):

src/pykiso/cli.py
TedRio: Actually Path(config_file).stem does exactly what I want and this whole function is useless 🤦‍♂️

Pog3k: this function will be then removed ?

src/pykiso/cli.py
Pog3k: yes pls no breaking change

src/pykiso/cli.py
TedRio: done

src/pykiso/cli.py
TedRio: done

src/pykiso/cli.py
TedRio: Done

src/pykiso/logging_initializer.py
TedRio: Well if I do this is_dir will return False if the folder does not exist and it will not be created

TedRio: but previous check is indeed useless, removed

src/pykiso/cli.py
```diff
@@ -204,9 +218,31 @@ def main(
...
 
+    log_index = 0
     for config_file in test_configuration_file:
+
         # Set the logging
-        logger = initialize_logging(log_path, log_level, verbose, report_type)
+        if log_paths:
+            yaml_name = get_file_name_without_extension(config_file)
+            # Put all logs in one file
+            if len(log_paths) == 1:
```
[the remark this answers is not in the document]
TedRio: Added the enumerate, for the check before the looping not sure it would make the code cleaner

src/pykiso/logging_initializer.py
sebclrsn: Absolutely right!
You could still flatten this a bit by doing:
```
if log_path.suffix == "":
    log_path.mkdir(parents=True, exist_ok=True)
```
which would already take care of the additional is_dir condition

Pog3k (approved):

src/pykiso/logging_initializer.py
TedRio: Done

PR #265, opened 2023-03-31
feat: allow proxy auto start and stop

sebclrsn:
This implements auto start & stop capabilities for the ProxyAuxiliary:
- when the first CCProxy is opened and the ProxyAuxiliary isn't started yet, start it
- when the last CCProxy is closed and the ProxyAuxiliary is running, stop it
- if all auxiliaries are marked with `auto_start: False` in the config, apply this to the created ProxyAuxiliary as well

-> Goal is to entirely hide the proxy from the user as he won't be able to access it anymore in case of an automatic proxy creation

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/PyCQA/isort.git
-    rev: 5.8.0
+    rev: 5.12.0
     hooks:
       - id: isort
         args: ["--profile", "black", "--line-length", "79"]
--- a/src/pykiso/lib/auxiliaries/proxy_auxiliary.py
+++ b/src/pykiso/lib/auxiliaries/proxy_auxiliary.py
@@ -92,8 +92,38 @@ class ProxyAuxiliary(DTAuxiliaryInterface):
             is_proxy_capable=True, tx_task_on=False, rx_task_on=True, **kwargs
         )
         self.channel = com
+        self._open_count = 0
         self.logger = self._init_trace(activate_trace, trace_dir, trace_name)
         self.proxy_channels = self.get_proxy_con(aux_list)
+
+    @property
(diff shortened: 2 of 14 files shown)


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/PyCQA/isort.git
-    rev: 5.8.0
+    rev: 5.12.0
```
sebastianpfischer, line 3: I think the docs also need some update no?

src/pykiso/lib/auxiliaries/proxy_auxiliary.py
```diff
@@ -92,9 +92,42 @@ def __init__(
...
+            communication channel closed.
+            When the first connection is opened and the counter is increased
+            from 0 to 1, the ``ProxyAuxiliary`` will be started and the attached
+            'physical' communication channel opened.
+        """
+        with self.lock:
+            return self._open_count
+
+    @_open_connections.setter
+    def _open_connections(self, value: int):
```
sebastianpfischer, line 120: Would a counting-semaphore not do the trick?

sebastianpfischer (changes requested):

Pog3k (approved):

src/pykiso/lib/auxiliaries/proxy_auxiliary.py
sebclrsn: Checking when `proxy.open_connections.acquire(blocking=False)` returns False seems way more obscure to me than having an explicit counter and some boolean operations

.pre-commit-config.yaml
sebclrsn: Clarified - no need to specify the pre-commit hook versions in the docs :)

sebastianpfischer (approved):

PR #282, opened 2023-05-05
feat: step report handler errors raised

yannpoupon:
Add unexpected errors traceback under the table created by the step report and fails the the test if an error was raised in the test.


Pog3k (approved):

Pog3k:
Pls change commit message because changelog is auto generated.
something like
fix: wrong step report resule

sebclrsn (approved):

PR #266, opened 2023-03-31
Feature/trc consecutive timestamps when merge

TedRio:
***** This implementation is far from perfect as not guaranteed to work with file version != from 2.0 / 2.1 *****
It's a temporary solution while waiting for this PR: https://github.com/hardbyte/python-can/pull/1530

PR #293, opened 2023-05-08
bugfix[codecov]: revert container version

sebastianpfischer:
I reverted the code in #266 that broke the codecov github job

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -18,13 +18,22 @@ jobs:
   # This workflow contains a single job called "build"
   build:
     # The type of runner that the job will run on
-    runs-on: eclipse/kiso-build-env:v0.2.1
+    runs-on: ubuntu-latest
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
 
+      # Pre requisite installation
+      - name: Prerequisite installation
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
+
+      - name: Install poetry
+        run: pip3.7 install poetry
+    
       # Runs the preparation
       - name: Prepare working environement
         run: poetry install --all-extras
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -18,7 +18,7 @@ jobs:
   # This workflow contains a single job called "build"
   build:
     # The type of runner that the job will run on
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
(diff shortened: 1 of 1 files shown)


Pog3k:
I changed this because our unittest  will fail with the "ubuntu" container. 
The unittest creates some test files with different timestamps in the meta file information.
In python we create a list of files which should actually be sorted by date but it doesn't work with the ubuntu image. 
Somehow it cant handle the date.
In wsl it works with an ubuntu.

Pog3k (changes requested):
maybe try other lightweight image .. debian ?

sebastianpfischer:
> maybe try other lightweight image .. debian ?

@Pog3k , I think we forgot to read the documentation -> https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

Our only possibilities are:
* `windows-latest` or `windows-2022`, `windows-2019`
* ` ubuntu-latest` or `ubuntu-22.04`
* `ubuntu-20.04` (with less runners)
* `macos-13` or `macos-13-xl`, `macos-latest`, `macos-12`, `macos-latest-xl` or `macos-12-xl`, `macos-11``

So only Ubunto-22.04 is here a possibility for us.
Do I forget something?

Pog3k:
Testet it on my pc with ubuntu:latest and python3.10 and it works.. can your try to use 3.10 instead of 3.7

.github/workflows/codecov.yml
```diff
@@ -18,13 +18,22 @@ jobs:
...
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
 
+      # Pre requisite installation
+      - name: Prerequisite installation
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
```
Pog3k, line 32: `python-version: '3.10`

sebastianpfischer:
@Pog3k, did you also test it with 3.8?

.github/workflows/codecov.yml
sebastianpfischer: Yea ok, I probably need to maintain the compatibility with 3.7 via the Eclipse Jenkins right?

Pog3k:
i updated it and codecov seems now working. 

So some weird behaviour with ubuntu and the old python 3.7 which is actually discontinued for the latest ubuntu...

Pog3k (approved):

flo117 (approved):

sebastianpfischer:
@Pog3k , thanks, have you seen: https://app.codecov.io/gh/eclipse/kiso-testing/commit/96b1b4703386b6b7a2a98bedf1f156c6df4f859a?

PR #298, opened 2023-05-24
fix: use open from pathlib instead of the builtin

sebclrsn:


Pog3k:
THIS IS JUST A QUICKFIX. 

Proper Teardown is already planned

Pog3k (approved):

PR #303, opened 2023-06-01
feat: change logger class for the test

yannpoupon:
Add the possibility to change the logger class with the CLI.

### DIFF
--- a/docs/advanced_usage/how_to_change_logger_class.rst
+++ b/docs/advanced_usage/how_to_change_logger_class.rst
@@ -37,4 +37,4 @@ See the following example :
 
 .. code:: bash
 
-  pykiso -c my_config.yaml --logger 'package_test.test_logger.TestLogger(str_arg=str_value,int_arg=12)'
+  pykiso -c my_config.yaml --logger 'package_test.test_logger.TestLogger(str_arg="str_value",int_arg=12)'
--- a/docs/advanced_usage/how_to_change_logger_class.rst
+++ b/docs/advanced_usage/how_to_change_logger_class.rst
@@ -1,0 +1,40 @@
+.. _change_logger_class:
+
+
+How to change the class of the logger
+-------------------------------------
+
+Change logger class
+~~~~~~~~~~~~~~~~~~~
+
+You can change the class of the loggers used in pykiso with a custom logger class.
+
+For example you have created a logger class (TestLogger) in the package called package_test in a file called logger_test.
+To use it you have pass the path to import the class as the value for --logger : package_test.test_logger.TestLogger
+
+.. code:: bash
+
+  pykiso -c my_config.yaml --logger package_test.test_logger.TestLogger
+
+Class with argument
+~~~~~~~~~~~~~~~~~~~
+
+If you have created a logger class that takes more argument than name and level
+to be initialized, you can specify them so that all loggers will be initialized
+with the same value.
+
+See the following example :
+
+.. code:: python
+
+   class TestLogger(logging.Logger):
+        def __init__(
+            self, name: str, str_arg: str, int_arg: int, level=0
+        ) -> None:
+            super().__init__(name, level)
+            self.str_arg = str_arg
+            self.int_arg = int_arg
+
+.. code:: bash
+
+  pykiso -c my_config.yaml --logger 'package_test.test_logger.TestLogger(str_arg=value,int_arg=12)'
--- a/docs/advanced_usage/how_to_change_logger_class.rst
+++ b/docs/advanced_usage/how_to_change_logger_class.rst
@@ -37,4 +37,4 @@ See the following example :
 
 .. code:: bash
 
-  pykiso -c my_config.yaml --logger 'package_test.test_logger.TestLogger(str_arg=value,int_arg=12)'
+  pykiso -c my_config.yaml --logger 'package_test.test_logger.TestLogger(str_arg=str_value,int_arg=12)'
--- a/src/pykiso/logging_initializer.py
+++ b/src/pykiso/logging_initializer.py
@@ -18,14 +18,24 @@ Integration Test Framework
 .. currentmodule:: logging
 
 """
+import importlib
+import json
 import logging
 import sys
 import time
+from functools import partialmethod
 from pathlib import Path
-from typing import List, NamedTuple, Optional
+from typing import Any, List, NamedTuple, Optional
 
 from .test_setup.dynamic_loader import PACKAGE
 from .types import PathType
+
+LEVELS = {
+    "DEBUG": logging.DEBUG,
+    "INFO": logging.INFO,
+    "WARNING": logging.WARNING,
+    "ERROR": logging.ERROR,
+}
 
 
 class LogOptions(NamedTuple):
@@ -37,6 +47,21 @@ class LogOptions(NamedTuple):
     log_level: str
     report_type: str
     verbose: bool
+
+
+class InternalLogsFilter(logging.Filter):
+    def filter(self, record):
+        """Filters internal log levels
+
+        :param record: event being logged
+
+        :return: False if internal logging, True otherwise
+        """
+        return record.levelno not in (
+            logging.INTERNAL_WARNING,
+            logging.INTERNAL_INFO,
+            logging.INTERNAL_DEBUG,
+        )
 
 
 # used to store the selected logging options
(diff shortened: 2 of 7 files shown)


src/pykiso/logging_initializer.py
```diff
@@ -234,3 +239,94 @@ def initialize_loggers(loggers: Optional[List[str]]) -> None:
...
+
+def import_object(path: str) -> Any:
+    """return the object based on the path.
+        For example : logging.Logger will return Logger
+
+    :param path: path to the object
+    """
+    if path:
+        components = path.split(".")
+        mod = importlib.import_module(components[0])
```
sebclrsn, line 252: You could also make use of the package kwarg:
```
components = path.split(".")
if len(components) == 1:
    return importlib.import_module(components[0])
else:
    return importlib.import_module(components[-1], package=component[:-1])

File: src/pykiso/logging_initializer.py [hunk 2]
```diff
@@ -234,3 +239,94 @@ def initialize_loggers(loggers: Optional[List[str]]) -> None:
...
+    # Get the argument to initialize the logger if needed
+    kwargs_log = {}
+    if "(" in logger:
+        logger_class, arg_logger = logger.split("(")
+        arg_logger = arg_logger[:-1].split(",")
+        kwargs_log = {arg.split("=")[0]: arg.split("=")[1] for arg in arg_logger}
+        # We try to load item to handle int, bool argument
+        for name, arg in kwargs_log.items():
+            try:
+                kwargs_log[name] = json.loads(arg.lower())
```
sebclrsn, line 305: It could be wiser to load the entire dict directly and raise if it fails instead of silently dropping the argument

docs/advanced_usage/how_to_change_logger_class.rst
```diff
@@ -0,0 +1,40 @@
...
+        def __init__(
+            self, name: str, str_arg: str, int_arg: int, level=0
+        ) -> None:
+            super().__init__(name, level)
+            self.str_arg = str_arg
+            self.int_arg = int_arg
+
+.. code:: bash
+
+  pykiso -c my_config.yaml --logger 'package_test.test_logger.TestLogger(str_arg=value,int_arg=12)'
```
Pog3k, line 40: I dont get how this should work .
actuall this part ->  str_arg=value

Is value a variable here  or should it be
str_arg="value"   ??

src/pykiso/logging_initializer.py
```diff
@@ -39,6 +49,21 @@ class LogOptions(NamedTuple):
     verbose: bool
 
 
+class InternalLogsFilter(logging.Filter):
+    def filter(self, record):
```
Pog3k, line 53:
```suggestion
    def filter(self, record)-> bool:
```

Pog3k: record type ?

src/pykiso/logging_initializer.py
```diff
@@ -234,3 +239,94 @@ def initialize_loggers(loggers: Optional[List[str]]) -> None:
     loggers_to_deactivate = set(relevant_loggers) - set(active_loggers)
     for logger_name in loggers_to_deactivate:
         logging.getLogger(logger_name).setLevel(logging.WARNING)
+
+
+def import_object(path: str) -> Any:
+    """return the object based on the path.
+        For example : logging.Logger will return Logger
+
+    :param path: path to the object
```
Pog3k, line 248: dont we need :return: for sphinx ? i know that your current description describes the return value

src/pykiso/logging_initializer.py
```diff
@@ -234,3 +239,94 @@ def initialize_loggers(loggers: Optional[List[str]]) -> None:
...
+    def wrapper(self, *arg, **kwargs):
+        func(self, *arg, **kwargs)
+        for handler in self.handlers:
+            self.removeHandler(handler)
+
+    return wrapper
+
+
+def change_logger_class(log_level: str, verbose: bool, logger: str):
+    """Change the class of all the logger of pykiso.
```
Pog3k, line 291: It also change the sublevels of the logger right ? Because you call it only once..

src/pykiso/logging_initializer.py
```diff
@@ -234,3 +239,94 @@ def initialize_loggers(loggers: Optional[List[str]]) -> None:
...
+
+    :param log_level: level of the log
+    :param logger: str of the path to the logger class
+    """
+    # Get the argument to initialize the logger if needed
+    kwargs_log = {}
+    if "(" in logger:
```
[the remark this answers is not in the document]
yannpoupon: Done

docs/advanced_usage/how_to_change_logger_class.rst
yannpoupon: Here value is just a str since it is already inside ' '.

src/pykiso/logging_initializer.py
yannpoupon: Done

src/pykiso/logging_initializer.py
yannpoupon: Done

src/pykiso/logging_initializer.py
yannpoupon: `json.loads` does not support dict, and the goal here is to transform the str given by the user to the appropriate type : 
'12' -> 12 
So if it fails it means it is just a str. The argument is not dropped he keeps the same value as given by the user.

src/pykiso/logging_initializer.py
yannpoupon: What I want in the end is not a module but an Object.

sebclrsn: Then a single getattr should be sufficient

src/pykiso/logging_initializer.py
sebclrsn: You could then consider using `ast.literal_eval` instead, which will allow strings passed as `my_str_param='str'` and `my_str_param="str"` as well as any other builtin python type

src/pykiso/logging_initializer.py
yannpoupon: Done

src/pykiso/logging_initializer.py
yannpoupon: Done

sebclrsn (approved):

TedRio (approved):

flo117 (approved):

flo117 (approved):

docs/advanced_usage/how_to_change_logger_class.rst
```diff
@@ -0,0 +1,40 @@
+.. _change_logger_class:
+
+
+How to change the class of the logger
```
sebastianpfischer, line 4: Why would it be needed? Which use-case did we accounted?

src/pykiso/logging_initializer.py
```diff
@@ -234,3 +241,96 @@ def initialize_loggers(loggers: Optional[List[str]]) -> None:
     loggers_to_deactivate = set(relevant_loggers) - set(active_loggers)
     for logger_name in loggers_to_deactivate:
         logging.getLogger(logger_name).setLevel(logging.WARNING)
+
+
+def import_object(path: str) -> Any:
```
sebastianpfischer, line 246: This feels a bit strange. Do we really want it to import any object? Seems like a security leak somehow. We are in the `logging_initializer.py`. Please make sure we only import the exact needed logging object.

sebastianpfischer (changes requested):
It would be nice to have a walk through together.

docs/advanced_usage/how_to_change_logger_class.rst
yannpoupon: The use case is to use kiso testing with fluentd to use structured logger, for this a logger using : https://github.com/fluent/fluent-logger-python#python-logginghandler-interface can be used.
So the log of the test could be send with the handler of the new Logger Class just by calling `pykiso --logger `

src/pykiso/logging_initializer.py
yannpoupon: A check  to see if the object imported is a subclass of `logging.Logger` has been implemented.

sebastianpfischer: It is better, I think you just need to update the return value to `Union[None, logging.Logging]`

sebastianpfischer (changes requested):

src/pykiso/logging_initializer.py
yannpoupon: Done

sebastianpfischer (approved):

PR #314, opened 2023-06-21
fix[readthedocs]: update python version

sebastianpfischer:


sebastianpfischer:
I'm not touching code as such so I guess the codecov can be ignored

TedRio (approved):

PR #332, opened 2023-07-21
275 use official trc reader

TedRio:


src/pykiso/lib/connectors/trc_handler.py
```diff
@@ -0,0 +1,325 @@
...
+        """Pick message format from file version and format initial message
+
+        :return: message informations formated in a string
+        """
+        # Fix error from python can -> message number should start at one not zero
+        self.msgnr = 1
+
+        if self.file_version == TRCFileVersion.V1_0:
+            self._msg_fmt_string = self.FORMAT_MESSAGE_V1_0
+            self._format_message = self._format_message_by_format
```
sebclrsn, line 229: Wouldn't it be way more concise to just run these lines once at the very end of the function?
```python
self._format_message = self._format_message_by_format
return self._format_message_by_format(msg, channel)
```

src/pykiso/lib/connectors/trc_handler.py
```diff
@@ -0,0 +1,325 @@
...
+        if msg.type not in ["EC", "ER", "ST"]:
+            if msg.is_extended_id:
+                arb_id = f"{msg.arbitration_id:07X}"
+            else:
+                arb_id = f"{msg.arbitration_id:04X}"
+        else:
+            arb_id = f"{msg.arbitration_id}"
+        data = [f"{byte:02X}" for byte in msg.data]
+
+        """For the time python-can was doing substraction with the first message timestamp
```
sebclrsn, line 256: Please make regular comments out of this, it quite confusing to have a docstring-like text block in the middle of a function IMO :D

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -382,10 +387,24 @@ def _cc_receive(
...
+
+        with trace.open("r") as trc:
+            data = trc.read().splitlines(True)
+            for line in data:
+                if line.startswith(";"):
+                    header_data.append(line)
+        return header_data
 
+    def _merge_trc(self) -> None:
+        """Merge all traces file in one and fix potential inconsitancies."""
```
sebclrsn, line 407:
```suggestion
        """Merge all traces file in one and fix potential inconsitencies."""
```

src/pykiso/lib/connectors/trc_handler.py
TedRio: Done

src/pykiso/lib/connectors/trc_handler.py
TedRio: Done

src/pykiso/lib/connectors/cc_pcan_can.py
TedRio: Done

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -399,115 +418,87 @@ def _merge_trc(self) -> None:
...
 
-                    # Fix message number and time offset inconstistancies
-                    corrected_data = CCPCanCan._fix_merge_inconsistencies(
-                        data[first_message_line:], offset, message_idx
-                    )
-                    message_idx += len(data[first_message_line:])
+            # Write header in result trace
+            with result_trace.open("w") as trc:
+                for line in data:
+                    if line[0] == ";":
```
sebclrsn, line 441:
```suggestion
                    if line.startswith(";"):
```

sebclrsn: Wouldn't it also be better to break as soon as the ';'  is no longer encountered? It would avoid going through the entire trc just to extract the header

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -382,10 +387,24 @@ def _cc_receive(
...
-    # TODO: refactor to use trc reader when new version of python can is release
-    def _merge_trc(self) -> None:
-        """Merge multiple trc files in one."""
+    @staticmethod
+    def _extract_header(trace: Path):
+        """Extract data from trc file header
+        
+        :param trace: trc file from which to extract the data
+        
+        :return: header data"""
```
sebclrsn, line 396:
```suggestion
        :return: header data
        """
```

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -382,10 +387,24 @@ def _cc_receive(
             log.exception(f"encountered error while receiving message via {self}")
             return {"msg": None}
 
-    # TODO: refactor to use trc reader when new version of python can is release
-    def _merge_trc(self) -> None:
-        """Merge multiple trc files in one."""
+    @staticmethod
+    def _extract_header(trace: Path):
```
sebclrsn, line 391:
```suggestion
    def _extract_header(trace: Path) -> List[str]:
```

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -382,10 +387,24 @@ def _cc_receive(
...
+        
+        :param trace: trc file from which to extract the data
+        
+        :return: header data"""
+        header_data = []
+
+        with trace.open("r") as trc:
+            data = trc.read().splitlines(True)
+            for line in data:
+                if line.startswith(";"):
```
sebclrsn, line 402: Same here, if the header data stops as soon as there is no ; start character anymore, why continue through the rest of the file?

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -399,115 +418,87 @@ def _merge_trc(self) -> None:
...
-        """
-        # Search for the start time in the header of the tace
-        for line in data[:first_message_line]:
-            if line.find(";$STARTTIME=") != -1:
-                start_time = float(line[len(";$STARTTIME=") :])
-                return CCPCanCan._convert_peak_format_to_start_time(start_time)
-        raise ValueError("No start time was found in the trace file")
+        messages = []
+        for trc in list_of_traces:
+            log.internal_debug(f"Merging trace {trc.name} into {result_trace.name}")
```
sebclrsn, line 467:
```suggestion
            log.internal_debug("Merging trace %s into %s", trc.name, result_trace.name)
```

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -399,115 +418,87 @@ def _merge_trc(self) -> None:
...
+                    if line[0] == ";":
+                        trc.write(line)
 
-                with result_trace.open("a") as merged_trc:
-                    merged_trc.writelines(corrected_data)
-                os.remove(file)
+            # Write all messages in result trace
+            with TRCWriterCanFD(result_trace) as writer:
+                writer.file_version = self.trc_file_version
+                log.internal_debug(f"TRC file version is {self.trc_file_version}")
```
sebclrsn, line 447:
```suggestion
                log.internal_debug("TRC file version is %s", self.trc_file_version)
```

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -399,115 +418,87 @@ def _merge_trc(self) -> None:
...
-                    # Fix message number and time offset inconstistancies
-                    corrected_data = CCPCanCan._fix_merge_inconsistencies(
-                        data[first_message_line:], offset, message_idx
-                    )
-                    message_idx += len(data[first_message_line:])
+            # Write header in result trace
+            with result_trace.open("w") as trc:
+                for line in data:
+                    if line[0] == ";":
+                        trc.write(line)
```
sebclrsn, line 442:
```suggestion
            result_trace.write_text(header_data)
```
Wouldn't this do exactly the same?

sebclrsn (changes requested):

src/pykiso/lib/connectors/cc_pcan_can.py
```diff
@@ -399,115 +421,80 @@ def _merge_trc(self) -> None:
...
-            # End of the trace header
-            first_message_line = 33
 
-            # Get start time of the first trc file
-            with list_of_traces[0].open("r") as trc:
-                data = trc.read().splitlines(True)
-
-                first_trc_start_time = CCPCanCan._get_trace_start_time(
-                    data, first_message_line
+                list_of_traces[0] = Path(
```
sebclrsn, line 427: Maybe I'm wrong, but this seems to be unnecessary as the content is overwritten afterwards, to me this does:
- rename the first trc to the expected name
- overwrite the first trc with the first trc's header, content (?) and the content of the other trcs

sebclrsn (approved):

src/pykiso/lib/connectors/trc_handler.py
```diff
@@ -0,0 +1,322 @@
...
+
+class TypedMessage(Message):
+    """Message with type attribut added and that can handle empty arbitration id.
+    Necessary because python-can does not handle types from CAN FD and informations
+    contained in Message are not enough to deduce it.
+    Some of those types (ex: ST) also have no arbitration id which cannot be handle by
+    python-can Message object.
+    """
+
+    def __init__(self, type: str, msg: Message):
```
stefandtotev, line 46: I don't know the details of the Message class, but it seems strange to use an object of the parent class to initialize the child class. Maybe since you're using the attributes of the parent class you can pass them like creating an object of type Message, just including the type?

src/pykiso/lib/connectors/trc_handler.py
```diff
@@ -0,0 +1,322 @@
...
+        self.is_rx = msg.is_rx
+        self.bitrate_switch = msg.bitrate_switch
+        self.error_state_indicator = msg.error_state_indicator
+        self._check = msg._check
+
+    def __repr__(self) -> str:
+        """String representation that can handle messages with no arbitration id
+
+        return: string representation of the message"""
+        if self.type not in ["EC", "ER", "ST"]:
```
stefandtotev, line 72: This combination of types seems to be used in multiple places, maybe it can be extracted into a constant or a class variable?

src/pykiso/lib/connectors/trc_handler.py
```diff
@@ -0,0 +1,322 @@
...
+                f"timestamp={self.timestamp}",
+                f"arbitration_id={self.arbitration_id:#x}",
+                f"is_extended_id={self.is_extended_id}",
+            ]
+        else:
+            args = [
+                f"timestamp={self.timestamp}",
+                f"arbitration_id={self.arbitration_id}",
+                f"is_extended_id={self.is_extended_id}",
+            ]
```
stefandtotev, line 83: Since the arbitration_id is the only difference, maybe this can be shortened, something like:
`args = [
        f"timestamp={self.timestamp}",
        f"arbitration_id={self.arbitration_id if self.type in ["EC", "ER", "ST"] else self.arbitration_id:#x}",
        f"is_extended_id={self.is_extended_id}",
]`

src/pykiso/lib/connectors/trc_handler.py
```diff
@@ -0,0 +1,322 @@
...
+        :param cols: list of columns to parse
+
+        :return: typed message or None if message type is not supported
+        """
+        dtype = cols[self.columns["T"]]
+        cols = self._nomalize_cols(cols)
+
+        if dtype in ["DT", "FD", "FB", "EC", "ER", "ST"]:
+            return self._parse_msg_V2_x(cols)
+        else:
```
stefandtotev, line 145: I don't think you need the else statement here

src/pykiso/lib/connectors/trc_handler.py
```diff
@@ -0,0 +1,322 @@
...
+        """
+        # Fix error from python can -> message number should start at one not zero
+        self.msgnr = 1
+
+        if self.file_version == TRCFileVersion.V1_0:
+            self._msg_fmt_string = self.FORMAT_MESSAGE_V1_0
+        elif self.file_version == TRCFileVersion.V2_0:
+            self._msg_fmt_string = self.FORMAT_MESSAGE_V2_0
+        elif self.file_version == TRCFileVersion.V2_1:
+            self._msg_fmt_string = self.FORMAT_MESSAGE_V2_1
```
stefandtotev, line 232: Maybe you can create a trc_file_version_to_format_message mapping between TRCFileVersion and FORMAT_MESSAGE and just do something like:
self._msg_fmt_string = self.trc_file_version_to_format_message[self.file_version]
This mapping can be reused in the if statement in the write_header function as well

src/pykiso/lib/connectors/trc_handler.py
```diff
@@ -0,0 +1,322 @@
...
+            serialized = self._msg_fmt_string.format(
+                msgnr=self.msgnr,
+                time=(msg.timestamp - self.first_timestamp) * 1000,
+                type=msg.type,
+                channel=channel,
+                id=arb_id,
+                dir="Rx" if msg.is_rx else "Tx",
+                dlc=msg.dlc,
+                data=" ".join(data),
+            )
```
stefandtotev, line 276: A lot of the parameters seem to be repeated, maybe you can do something like:
`
format_fields = {
        "msgnr": self.msgnr,
        "time": (msg.timestamp - self.first_timestamp) * 1000,
        "channel": channel,
        "id": arb_id,
        "dir": "Rx" if msg.is_rx else "Tx",
        "dlc": msg.dlc,
        "data": " ".join(data),
}
if self.file_version != TRCFileVersion.V1_0:
    format_fields["type"] = (msg.timestamp - self.first_timestamp) * 1000

return self._msg_fmt_string.format(**format_fields)
`

src/pykiso/lib/connectors/cc_pcan_can.py
sebclrsn: Resolved, this is done to ensure that all of the other traces are deleted

src/pykiso/lib/connectors/trc_handler.py
sebclrsn: Definitely, I switched to proper inheritance instead :)

src/pykiso/lib/connectors/trc_handler.py
sebclrsn: Added as constant `NO_ARBITRATION_ID_TYPE` ;)

src/pykiso/lib/connectors/trc_handler.py
sebclrsn: Good suggestion, but I think we'll focus on readability and leave it as it is it that's fine for you

src/pykiso/lib/connectors/trc_handler.py
sebclrsn: Definitely, applied!

Pog3k (approved):

PR #335, opened 2023-07-31
Improve log for uds auxiliary for performance

yannpoupon:


src/pykiso/test_coordinator/test_case.py
```diff
@@ -349,3 +349,22 @@ def func_wrapper(self: BasicTest) -> None:
         return func_wrapper
 
     return decorator
+
+
+def disable_logging(level: int = logging.CRITICAL):
```
sebclrsn, line 354: This would have expected to have this in logging_initializer

src/pykiso/lib/auxiliaries/udsaux/uds_auxiliary.py
```diff
@@ -132,14 +134,14 @@ def send_uds_raw(
...
-        resp = UdsResponse(resp)
-        return resp
+        if log.isEnabledFor(logging.getLogger().level):
+            resp_print = (
+                f"UDS response received {['0x{:02X}'.format(i) for i in resp]}"
+                if not isinstance(resp, bool)
+                else resp
+            )
+            log.internal_info(resp_print)
+        return UdsResponse(resp)
```
sebclrsn, line 144: Wouldn't it make more sense to have a proper `__repr__` implemented in `UdsResponse`? 
You would then just do `log.internal_info(resp)`

Would be nice to have something like: 
```python
def __repr__(self):
    if self.is_negative:
        return f"NegativeUdsResponse({bytes(self.data).hex(sep=' ')}, nrc={self.nrc.name}})"
    return f"{self.__class__.__name__}({bytes(self.data).hex(sep=' ')})"

File: src/pykiso/lib/auxiliaries/udsaux/uds_auxiliary.py [hunk 3]
```diff
@@ -114,9 +114,11 @@ def send_uds_raw(
             False
         """
         try:
-            log.internal_info(
-                f"UDS request to send '{['0x{:02X}'.format(i) for i in msg_to_send]}'"
-            )
+            if log.isEnabledFor(logging.getLogger().level):
```
sebclrsn, line 117: IMO this should be tackled in a separate task to have a proper filtering of our internal log levels as for any other default log level

src/pykiso/lib/auxiliaries/udsaux/uds_auxiliary.py
```diff
@@ -193,12 +195,13 @@ def send_uds_config(
...
             if req_resp_data is None:
                 req_resp_data = True
-            log.internal_info(f"UDS response received {req_resp_data}")
+            log.internal_info("UDS response received %s", req_resp_data)
             return req_resp_data
         except AttributeError:
             # Service not found, raised by getattr()
             log.exception(
-                f"Could not send UDS config request: unknown service {uds_service_name}"
+                "Could not send UDS config request: unknown service %s",
```
sebclrsn, line 203: It makes less sense to have percent formatting for warning and error levels as these are usually never filtered out

src/pykiso/lib/auxiliaries/udsaux/uds_auxiliary.py
Pog3k: How does the user turn off/on the internal logs ?

I would have thought that he can control it via the uds_aux level. Or via decorator ?

src/pykiso/lib/auxiliaries/udsaux/uds_auxiliary.py
yannpoupon: It has been reverted

src/pykiso/lib/auxiliaries/udsaux/uds_auxiliary.py
yannpoupon: Done

src/pykiso/test_coordinator/test_case.py
yannpoupon: Done

src/pykiso/lib/auxiliaries/udsaux/uds_auxiliary.py
yannpoupon: Actually internal logs can only be off if the user choose a level higher than INFO for log level of the test which is why I added a decorator `disable_logging` to be able to disable logging for a specific test

sebclrsn: The disable_logging is totally fine, it's more regarding the workarounds you had to do to avoid the internal log level messages to be evaluated
This should then be fixed in logging_initializer, where the internal log levels should always be lower than the non-internal ones so that the logging module can filter them out by itself

sebclrsn (approved):

src/pykiso/lib/auxiliaries/udsaux/common/uds_callback.py
```diff
@@ -294,15 +296,17 @@ def handle_data_download(
                 if data[0] != expected_pci:
                     block_data_len += len(data) - 1
                     log.internal_warning(
-                        f"Consecutive frame missed: "
-                        f"expected PCI {hex(expected_pci)}, got {hex(data[0])}"
+                        "Consecutive frame missed: expected PCI "
+                        f"{hex(expected_pci)}, got {hex(data[0])}",
```
Pog3k, line 300: stil f string ? was this intended ?

yannpoupon: Yes, Sebastian Clerson said : 
It makes less sense to have percent formatting for warning and error levels as these are usually never filtered out

And since there is an if condition you do not render the f string everytime but only when there is an error

PR #339, opened 2023-08-07
Add tags to global config

yannpoupon:


src/pykiso/global_config.py
```diff
@@ -135,15 +140,21 @@ def grab_cli_config(func) -> Callable:
...
+            :param click_context: click context
             :param args: positonal arguments
             :param kwargs: named arguments
             """
-            object_config = Grabber.create_config_object(kwargs)
+            kwargs_copy = copy(kwargs)
+            click_args = cli.eval_user_tags(click_context)
+            for tag in list(click_args.keys()):
+                click_args[tag.replace("-", "_")] = click_args.pop(tag)
+            kwargs_copy.update(click_args)
```
sebclrsn, line 155: You could also avoid the dict copy by merging both dicts into a new one:
`kwargs_with_tags = {**click_args, **kwargs}`

src/pykiso/global_config.py
```diff
@@ -135,15 +140,21 @@ def grab_cli_config(func) -> Callable:
...
 
+            :param click_context: click context
             :param args: positonal arguments
             :param kwargs: named arguments
             """
-            object_config = Grabber.create_config_object(kwargs)
+            kwargs_copy = copy(kwargs)
+            click_args = cli.eval_user_tags(click_context)
+            for tag in list(click_args.keys()):
+                click_args[tag.replace("-", "_")] = click_args.pop(tag)
```
sebclrsn, line 154:
```suggestion
            # replace all dashes with underscore to make valid variable names out of the tag names
            click_args = {tag_name.replace("-", "_"): tag_values for tag_name, tag_value in click_args.items())
```

tests/test_global_config.py
```diff
@@ -50,32 +51,32 @@
...
         report_type,
-        variant,
-        branch_level,
         pattern,
     ):
         return None
 
     config_parser()
+    context_mock = mocker.patch("click.Context")
+    context_mock.args = ["--variant", "variant1,", "--branch-level", "daily"]
```
sebclrsn, line 72: Where are the assertions? 🤔

sebclrsn (changes requested):

tests/test_global_config.py
yannpoupon: I just changed the implementation of the fixture to replace the fact that the tag and branch level were put as input parameters for the function which is not the case for pykiso CLI, so the test will be runned with the same logic as the cli.
But I added a new test.

src/pykiso/global_config.py
yannpoupon: Done

src/pykiso/global_config.py
yannpoupon: Done

TedRio (approved):

sebclrsn (approved):

PR #341, opened 2023-08-07
Bugfix/disable run command when auxiliary is stopped

sebclrsn:
- Raise an AuxiliaryNotStarted error if the user wants to execute a command while the auxiliary is stopped (otherwise it messes up the input queue of the auxiliary)
- Add the `timeout_result` keyword argument to `run_command` to be able to specify a command's return value in case of a timeout (for flexibility and future refactorings)

### DIFF
--- a/src/pykiso/interfaces/dt_auxiliary.py
+++ b/src/pykiso/interfaces/dt_auxiliary.py
@@ -27,7 +27,7 @@ import threading
 from enum import Enum, unique
 from typing import Any, Callable, List, Optional
 
-from ..exceptions import AuxiliaryCreationError
+from ..exceptions import AuxiliaryCreationError, AuxiliaryNotStarted
 from ..logging_initializer import add_internal_log_levels, initialize_loggers
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


TedRio (approved):

src/pykiso/interfaces/dt_auxiliary.py
```diff
@@ -96,6 +96,7 @@ def run_command(
         cmd_data: Any = None,
         blocking: bool = True,
         timeout_in_s: int = 5,
+        timeout_result: Any = None,
```
sebastianpfischer, line 99: Who should set this value? run_command can potentially be called by a tester script right?

sebastianpfischer (changes requested):

src/pykiso/interfaces/dt_auxiliary.py
sebclrsn: Absolutely, this just allows returning a different value than None on a timeout of the command, which removes any need to override run_command in case a more meaningful return value is wanted

Pog3k (approved):

PR #343, opened 2023-08-08
refactor: set internal log levels lower than non internal ones

sebclrsn:
**Tested and validated**

Suggestion to set the internal log levels lower than the non internal ones to that they are directly filtered out by the logging module if the configured log level is higher

### DIFF
--- a/src/pykiso/logging_initializer.py
+++ b/src/pykiso/logging_initializer.py
@@ -50,21 +50,6 @@ class LogOptions(NamedTuple):
     verbose: bool
 
 
-class InternalLogsFilter(logging.Filter):
-    def filter(self, record: logging.LogRecord) -> bool:
-        """Filters internal log levels
-
-        :param record: event being logged
-
-        :return: False if internal logging, True otherwise
-        """
-        return record.levelno not in (
-            logging.INTERNAL_WARNING,
-            logging.INTERNAL_INFO,
-            logging.INTERNAL_DEBUG,
-        )
-
-
 # used to store the selected logging options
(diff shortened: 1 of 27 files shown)


src/pykiso/logging_initializer.py
```diff
@@ -151,10 +137,10 @@ def initialize_logging(
...
-    if not verbose:
-        # filter internal log levels
-        stream_handler.addFilter(InternalLogsFilter())
+    # disable internal logs if no verbose is wanted
+    stream_handler.setLevel(
+        LEVELS[log_level] if not verbose else (LEVELS[log_level] - 1)
+    )
     root_logger.addHandler(stream_handler)
 
     root_logger.setLevel(LEVELS[log_level])
```
yannpoupon, line 146:
```suggestion
    root_logger.setLevel(LEVELS[log_level] - 1 if (verbose or log_path) else LEVELS[log_level] )
```
Else the log level will never be at internal level.

sebclrsn: Done ;)

src/pykiso/logging_initializer.py
```diff
@@ -317,9 +307,6 @@ def change_logger_class(log_level: str, verbose: bool, logger: str):
         logger_class.__init__ = partialmethod(
             logger_class.__init__, level=LEVELS[log_level], **kwargs_log
```
yannpoupon, line 308:
```suggestion
            logger_class.__init__, **kwargs_log
```
This level can be removed else if we change logger class the level will never be internal one for new log

PR #350, opened 2023-08-16
Refactor/internal log levels

sebclrsn:
Set the internal log levels lower than the non internal ones to that they are directly filtered out by the logging module if the configured log level is higher


src/pykiso/logging_initializer.py
```diff
@@ -317,9 +307,6 @@ def change_logger_class(log_level: str, verbose: bool, logger: str):
         logger_class.__init__ = partialmethod(
             logger_class.__init__, level=LEVELS[log_level], **kwargs_log
```
yannpoupon, line 308:
```suggestion
            logger_class.__init__, **kwargs_log
```
Remove level for new logger so only the root logger handle the logic level.

sebastianpfischer (approved):
Cannot test it but looks fine. How relevant is it to the users? If it is, please update `docs/whats_new/version_ongoing`

PR #365, opened 2023-09-07
feat(grpc_client): grpc as client

sebastianpfischer:
Please review,

I use the PR to get an input regarding the coverage report, on my Pc, the following tests are failing:
* test_logging_initializer
* test_text_result


sebastianpfischer:
@BKaDamien, @sebclrsn, @Pog3k please review?

src/pykiso/lib/connectors/cc_grpc_client.py
```diff
@@ -0,0 +1,216 @@
...
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0.
+#
+# SPDX-License-Identifier: EPL-2.0
+##########################################################################
+
+"""
+Communication Channel Via GRPC
+********************************
```
Pog3k, line 12:
```suggestion
******************************
```

src/pykiso/lib/connectors/cc_grpc_client.py
```diff
@@ -0,0 +1,216 @@
...
+        spec.loader.exec_module(self.grpc_datamodel)
+        # Import the generated grpc protobuf file
+        # Import the module
+        spec = importlib.util.spec_from_file_location(
+            Path(generated_protobuf_grpc_file).stem, Path(generated_protobuf_grpc_file)
+        )
+        self.grpc_stub_module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(self.grpc_stub_module)
+        # Extract the stubs
+        module_attributes = inspect.getmembers(self.grpc_stub_module, inspect.isclass)
```
Pog3k, line 85: what are you doing with inspect ? why do we need thos stubs ?

Pog3k:
i miss an example file.
There could be a simple python file which acts as a server which can be startet as external process.

Would it be possible to showcase a whole test ? -> Yaml config, Com auxiliary ?! testsuite + testfile

src/pykiso/lib/connectors/cc_grpc_client.py
```diff
@@ -0,0 +1,216 @@
...
+from typing import Dict, Optional
+
+import grpc
+
+from pykiso import connector
+
+log = logging.getLogger(__name__)
+
+
+class GrpcClient(connector.CChannel):
```
Pog3k, line 39: Not sure if this is an unwritten pykiso rule (:
```suggestion
class CCGrpcClient(connector.CChannel):
```

src/pykiso/lib/connectors/cc_grpc_client.py
```diff
@@ -0,0 +1,216 @@
...
+        self.channel.close()
+
+    def _cc_send(self, msg: bytes, **kwargs) -> None:
+        """Send message to grpc server.
+
+        :param msg: message to send, should bytes.
+
+        In addition to the message, the following keyword arguments are supported:
+
+            - rpc_name: name of the rpc to trigger
```
Pog3k, line 138: Why hide them in kwargs ? 
You could specify them in the method as optional which are defaulted to None.
-> Better code completion experience

src/pykiso/lib/connectors/cc_grpc_client.py
```diff
@@ -0,0 +1,216 @@
...
+            - message_name: name of the message to use
+            - message_attributes: attributes to use in dictionary format
+
+        """
+        # Extract needed parameters or the default once
+        service_name = kwargs.get("service_name", self.default_service_name)
```
sebclrsn, line 144: As these keywords arguments are extremely specific to this connector, I still have troubles to understand how this connector can be used by a generic auxiliary 
Maybe an example would help!

src/pykiso/lib/connectors/cc_grpc_client.py
```diff
@@ -0,0 +1,216 @@
...
+
+        :return: result of the rpc
+        """
+        try:
+            if message_attributes:
+                return rpc_method(message_class(**message_attributes))
+            else:
+                return rpc_method(message_class())
+        except Exception as e:
+            logging.error(f"Error while triggering rpc: {e}")
```
sebclrsn, line 201:
```suggestion
            log.exception(f"Error while triggering rpc: {e}")
```

src/pykiso/lib/connectors/cc_grpc_client.py
```diff
@@ -0,0 +1,216 @@
...
+            return None
+
+    def _cc_receive(self, timeout: float = 0.0000001) -> Dict[str, Optional[bytes]]:
```
sebclrsn, line 204:
```suggestion
    def _cc_receive(self, timeout: float = 1e-6) -> Dict[str, Optional[bytes]]:
```

src/pykiso/lib/connectors/cc_grpc_client.py
sebastianpfischer: https://grpc.io/docs/what-is-grpc/core-concepts/ -> search stub

src/pykiso/lib/connectors/cc_grpc_client.py
sebastianpfischer: because of https://kiso-testing.readthedocs.io/en/latest/advanced_usage/how_to_connector.html#parameters-and-return-values

src/pykiso/lib/connectors/cc_grpc_client.py
sebastianpfischer: https://kiso-testing.readthedocs.io/en/latest/advanced_usage/how_to_connector.html#parameters-and-return-values

I added an example but I did not test it yet, I will do that eventually (before we merge this PR. Nevertheless, I think you can take a look at it and tell me what you think.

I may need to extend the com-aux to support the feature I need. I thought it was already part of it

sebastianpfischer:
@Pog3k, @sebclrsn please take a look at the example, I will add also a server when I have the time so that I can fully test it.

Nevertheless, please provide me some feedback on it.

Pog3k (approved):

examples/test_suite_grpc/test_suite_grpc_client.py
```diff
@@ -0,0 +1,45 @@
...
+        answer = com_aux.send_message("")
+        assert answer == "Hello World!"
+
+
+@pykiso.define_test_parameters(suite_id=1, case_id=2, aux_list=[com_aux])
+class SpecificTest(pykiso.BasicTest):
+    """With specific parameters"""
+
+    def test_run(self):
+        answer = com_aux.send_message(
```
sebclrsn, line 38: How can this work? Our auxiliaries, including the CommunicationAuxiliary, don't accept any keyword args, so if we want to integrate this as a communication channel then we'll need to make all auxiliaries accept variable kwargs

I'm still somehow thinking that a GrpcClient would fit better with a simple auxiliary without cchannel

sebastianpfischer: Please take a look at the example now that it is working

sebastianpfischer:
@sebclrsn, please proceed with the review

sebclrsn:
> @sebclrsn, please proceed with the review
> 

Code looks good, we could theoretically merge it. 
Still, I'd like to know what the added value is in implementing the GRPC client within a communication channel, i.e. what an auxiliary holding that channel could add except for respecting the aux-channel architecture.

From what I see it would fit better as a GrpcClientAuxiliary, but maybe you have a different use case that explains this choice?

sebastianpfischer:
> > @sebclrsn, please proceed with the review
> 
> Code looks good, we could theoretically merge it. Still, I'd like to know what the added value is in implementing the GRPC client within a communication channel, i.e. what an auxiliary holding that channel could add except for respecting the aux-channel architecture.
> 
> From what I see it would fit better as a GrpcClientAuxiliary, but maybe you have a different use case that explains this choice?

@sebclrsn, I was also wondering if I should create an auxiliary instead but what convinced me to go with a connector was a use-case where you want to run different simulations behind the grpc client (or potentially in the future, a grpc server). If I had put it as aux, I would not have been able to simply add one or many additional logic here.

In addition, grpc is directly creating its "connection" too. Of course, we could have mock it , like for the uds but ip communication is really cheap now and the overhead we had for CAN is not worth here.

sebclrsn (approved):

PR #385, opened 2023-10-05
fix: enable internal file logging when junit option is provided

sebclrsn:
When running pykiso without verbose but with the junit option provided, the internal logs were no longer displayed in the provided log file


sebastianpfischer (approved):
cool

Pog3k (approved):

PR #386, opened 2023-10-05
fix: add event to avoid deadlocks on run_command calls

sebclrsn:
When a callback executed in _receive_message calls run_command while the auxiliary is being deleted, this run_command call never gets to acquire the lock (already acquired by delete_instance in the main thread) and _receive_message never terminates either.
This leads to deadlocks on delete_instance calls.

Solution is to add an event acting as a guard and preventing run_command to try and acquire the lock while the auxiliary is being deleted


sebastianpfischer (approved):

sebastianpfischer:
@TedRio, do you still want to review or do we merge it?

Pog3k:
Waiting for validation

PR #391, opened 2023-10-10
fix: support normal CAN for trc files

TedRio:


src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
```diff
@@ -491,5 +491,5 @@ def _remove_offset(
 
     def shutdown(self) -> None:
         """Destructor method."""
-        if self.logging_activated:
+        if self.logging_activated and self.is_fd:
```
sebclrsn, line 494: Is it really a viable solution to just disable trc merging for non-FD CAN?

sebastianpfischer: Where is the bug from? PeakCAN Lib?

Pog3k (approved):

src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
TedRio: It is not really a bug, just not supported in their read/write classes. It just mean that we do not handle the merging for some of the older version of trc format.  I updated the pull request so that only those formats are filtered.

TedRio: I checked with them, it is working as intended

src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
```diff
@@ -26,6 +26,8 @@
 from pathlib import Path
 from typing import Dict, List, Optional, Union
 
+from can import TRCFileVersion
```
sebclrsn, line 29: This should be put in the try block

sebclrsn (approved):

PR #403, opened 2023-11-07
fix: multiple execution if file pattern is provided

Pog3k:
Prevent multiple execution when -p flag is given through the cli. 

Happens when multiple suite_dir are given in the config file.


sebclrsn (commented):
Just add an entry in the what's new and we're good to go :)

Pog3k:
> Just add an entry in the what's new and we're good to go :)

Normally we just add new features there. This is just a bug fix ?!

sebclrsn (approved):

PR #427, opened 2023-12-11
feat: can auxiliary

IliyanKordev:


src/pykiso/can_parser.py
```diff
@@ -0,0 +1,73 @@
+##########################################################################
```
Pog3k, line 1: Use the eclipse header. Just check the other files in this repo

examples/test_suite_1/test_suite_1.py
```diff
@@ -162,6 +162,7 @@ def test_run(self):
         logging.info(f"------------resume auxiliaries run--------------")
         aux3.resume()
         aux2.resume()
+        print(aux4.get_message("BHU", ["KeyExchange_Master_BHU", "BHU_NetworkStatus"]))
```
Pog3k, line 165: Remove this line. 
In this repo always generic examples. Nothing from your workplace

Pog3k: Please leave the dummy example untouched in this context.. This example shall always run without external HW.

PR #429, opened 2023-12-11
feat: can auxiliary

IliyanKordev:


sebastianpfischer:
I'm confused by this PR, could you document a bit more what it is suppose to do? Maybe also add the code change related to the can-auxiliary to the PR.
Until then, I will close it and will be looking forward you reopening it :)

Issue #373, opened 2023-09-27
Can Auxiliary


Pog3k:
**in order to**  send can messages by dbc signal names
**we want to** create a CAN auxiliary 

Methods like send can message and cyclic can message.
receive by signal names ??

PR #432, opened 2023-12-14
(POC) feat: can aux

IliyanKordev:


src/pykiso/lib/auxiliaries/can_auxiliary.py
```diff
@@ -0,0 +1,296 @@
...
+            if isinstance(value, str):
+                signals[signal] = int.from_bytes(value.encode("utf8"), byteorder="big")
+        try:
+            message = self.parser.dbc.get_message_by_name(message)
+        except KeyError:
+            raise ValueError(f"{message} is not a message defined in the DBC file.")
+        msg_to_send = self.parser.encode(message, signals)
+        self.channel.cc_send(msg_to_send[0], remote_id=msg_to_send[1])
+
+    def run_command(
```
sebclrsn, line 227: Why is this API reimplemented? It's already part of the DTAuxiliaryInterface

src/pykiso/lib/auxiliaries/can_auxiliary.py
```diff
@@ -0,0 +1,296 @@
...
+                continue
+            for signal_name in expected_signals_names:
+                if last_can_msg.signals[signal_name] != expected_signals[signal_name]:
+                    continue
+            return last_can_msg
+        
+        return None
+
+    @staticmethod
+    def _is_payload_valid(payload: Optional[bytes]) -> bool:
```
sebclrsn, line 198: This doesn't look like it belongs here

src/pykiso/lib/auxiliaries/can_auxiliary.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def __init__(self, com: CChannel, dbc_file: str, **kwargs):
+        """Constructor.
+
+        :param com: CChannel that supports raw communication over CAN
+        """
+        super().__init__(
+            is_proxy_capable=True, tx_task_on=True, rx_task_on=True, **kwargs
+        )
+        self.tx_task_on = False
```
sebclrsn, line 76: This should be set in the line above

src/pykiso/lib/auxiliaries/can_auxiliary.py
```diff
@@ -0,0 +1,296 @@
...
+        self.recv_can_messages = {}
+        self.channel = com
+        self.queue_tx = queue.Queue()
+        self.queueing_event = threading.Event()
+        self.collect_messages = functools.partial(_collect_messages, com_aux=self)
```
sebclrsn, line 81: These 3 attributes don't seem to be used, you could remove the `_collect_messages` class above too

sebclrsn (changes requested):

src/pykiso/lib/auxiliaries/can_auxiliary.py
IliyanKordev: Ok, i will remove it

src/pykiso/lib/auxiliaries/can_auxiliary.py
IliyanKordev: ok

src/pykiso/lib/auxiliaries/can_auxiliary.py
IliyanKordev: Ok, i will remove it and test it

src/pykiso/can_message.py
```diff
@@ -0,0 +1,28 @@
...
+"""
+
+from typing import Any
+
+
+class CanMessage():
+    def __init__(self, name: str, signals: dict[str, Any], time_stamp: float) -> None:
+        self.name = name
+        self.signals = signals
+        self.time_stamp = time_stamp
```
sebclrsn, line 28:
```suggestion
        self.timestamp = timestamp
```

src/pykiso/can_message.py
```diff
@@ -0,0 +1,28 @@
...
+:synopsis: Message transfer by can
+
+.. currentmodule:: message
+"""
+
+from typing import Any
+
+
+class CanMessage():
+    def __init__(self, name: str, signals: dict[str, Any], time_stamp: float) -> None:
```
sebclrsn, line 25:
```suggestion
    def __init__(self, name: str, signals: dict[str, Any], timestamp: float) -> None:
```

sebclrsn, line 25: We are supposed to be compatible with python 3.7+, so using builtin types as type hints isn't possible

Use the ones from typing instead:
```
from typing import Any, Dict

Dict[str, Any]
```

src/pykiso/lib/auxiliaries/can_auxiliary.py
```diff
@@ -0,0 +1,209 @@
...
+                message_timestamp = float(rcv_data.get("timestamp", 0))
+                old_can_message = self.can_messages.get(message_name, None)
+                if old_can_message is None:
+                    can_msg = CanMessage(message_name, message_signals, message_timestamp)
+                    self.can_messages[message_name] = can_msg
+                else:
+                    self.can_messages[message_name].name = message_name
+                    self.can_messages[message_name].signals = message_signals
+                    self.can_messages[message_name].time_stamp = message_timestamp
+        except (Exception, KeyError):
```
sebclrsn, line 100: Simply catching the `Exception` will already catch the `KeyError`, if you need a different error message you might want to handle them separately:

```python
except KeyError:
    log.exception("Some message")
except Exception:
    log.exception("My very generic message")

File: src/pykiso/lib/auxiliaries/can_auxiliary.py [hunk 8]
```diff
@@ -0,0 +1,209 @@
...
+        timeout: float = 0.2
+    ) -> dict[str, any]:
+        """Get the last message with certain timeout in seconds.
+        :param message_name: name of the message to receive
+        :param timeout time to wait till a message receives in seconds
+
+        :return: list of last can messages or None if no messages for this component
+        """
+        last_msg = self.can_messages.get(message_name, None)
+        time.sleep(timeout)
```
sebclrsn, line 137: This isn't precise at all, it anyone wants to know how long it took until the message arrived he can't

Instead of having the potential message as `can_messages` values, you could also have a Queue with maxsize=1 which would allow you to use the non-polling APIs `.get(timeout=...)`

sebclrsn (changes requested):

src/pykiso/lib/auxiliaries/can_auxiliary.py
IliyanKordev: Ok, i will separate them

src/pykiso/can_message.py
IliyanKordev: ok, i will change it

src/pykiso/can_message.py
IliyanKordev: ok, i will change it

IliyanKordev: Ok, i will fix it

examples/test_can/test_can.py
```diff
@@ -0,0 +1,108 @@
...
+can_aux1: CanAuxiliary
+can_aux2: CanAuxiliary
+
+
+@pykiso.define_test_parameters(suite_id=1, case_id=1, aux_list=[can_aux1, can_aux2])
+class CanAuxTest(pykiso.BasicTest):
+    def setUp(self):
+        pass
+
+    def test_run(self):
```
Pog3k, line 38: Normally tests have an assert statement which you don't use here.. 
I would suggest to reduce the example to only one  extra thread.
Just start tx thread and wait for incoming message in this test method. 

Do an assert on the received message where you know the signal content

IliyanKordev: Ok, i agree, i will fix it

examples/test_can/test_can.py
```diff
@@ -0,0 +1,108 @@
...
+        send_t.start()
+        time.sleep(0.3)
+        recv_t.start()
+        send_t.join()
+        recv_t.join()
+
+        send_t = threading.Thread(
+            target=self.send_message, args=[can_aux1, messages_to_send, 0.2]
+        )
+        recv_t = threading.Thread(
```
Pog3k, line 68: don't start an extra receive thread. 
Just receive the message in this method an do an assert on the expected message

examples/test_can/test_can.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+
+        send_t.start()
+        time.sleep(0.3)
+        recv_t.start()
+        send_t.join()
+        recv_t.join()
+
+        send_t = threading.Thread(
+            target=self.send_message, args=[can_aux1, messages_to_send, 0.2]
```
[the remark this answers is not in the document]
IliyanKordev: ok, i will define it

examples/test_can/test_can.py
```diff
@@ -0,0 +1,108 @@
...
+        ]
+
+        send_t = threading.Thread(
+            target=self.send_message, args=[can_aux1, messages_to_send, 0.2]
+        )
+        recv_t = threading.Thread(
+            target=self.get_last_msg_and_wait_for_new_msg, args=[can_aux2, 1]
+        )
+
+        send_t.start()
```
Pog3k, line 59: I think here is a good point to create another test.

def test_wait_for_expected_signals(self):

IliyanKordev: Ok, i will separate them

examples/test_can/test_can.py
IliyanKordev: ok

examples/test_can/test_can.py
Pog3k, line 38: Every method which starts with "test"_ is considered to be run as a test from the unittest framework
I think you split the test_run into 2 tests 
Name the first
```suggestion
    def test_send_and_receive_message(self):
```
and the second one
def test_wait_for_expected_signals(self):

IliyanKordev: Ok, i will separate them

examples/test_can/test_can.py
```diff
@@ -0,0 +1,108 @@
...
+    def tearDown(self):
+        pass
+
+    def send_message(self, can_aux, messages_to_send, time_between_messages):
+        # send multiple messages one by one from can auxiliary with certain interval
+        for message_to_send in messages_to_send:
+            can_aux.send_message(message_to_send.name, message_to_send.signals)
+            time.sleep(time_between_messages)
+
+    def get_last_msg_and_wait_for_new_msg(self, can_aux, timeout):
```
Pog3k, line 88: a better name would have rx_thread  but as suggested just put the content in your test method  which is currently "test_run"

examples/test_can/test_can.py
```diff
@@ -0,0 +1,108 @@
...
+        send_t.start()
+        time.sleep(0.3)
+        recv_t.start()
+        send_t.join()
+        recv_t.join()
+
+    def tearDown(self):
+        pass
+
+    def send_message(self, can_aux, messages_to_send, time_between_messages):
```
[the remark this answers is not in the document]
IliyanKordev: ok, i agree

examples/test_can/test_can.py
IliyanKordev: ok

sebclrsn (approved):

Pog3k (approved):

sebastianpfischer:
Realise https://github.com/eclipse/kiso-testing/issues/373
Will be refactored with https://github.com/eclipse/kiso-testing/issues/452

PR #444, opened 2024-02-05
fix: fix log level level with pytest

TedRio:


sebclrsn (approved):
Thanks, that one was tricky :D 

You could apply the same fix for the FileHandler, we never know...

sebastianpfischer (approved):

Pog3k (approved):

PR #459, opened 2024-02-28
fix: fix timestamp inconsistencies between Windows and Linux

TedRio:


src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
```diff
@@ -34,6 +34,16 @@
...
+try:
+    # use to avoid timestamp inconsistances between pykiso and python can
+    import uptime
+
+    if uptime.boottime() is None:
+        boottime_epoch = 0
+    else:
+        boottime_epoch = uptime.boottime().timestamp()
+except ImportError:
+    boottime_epoch = 0
```
sebastianpfischer, line 46:
```suggestion
boottime_epoch = 0
try:
    # use to avoid timestamp inconsistances between pykiso and python can in linux
    import uptime
    if not uptime.boottime():
        boottime_epoch = uptime.boottime().timestamp()
except:
    pass
```

sebastianpfischer (approved):

Pog3k (approved):

src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
sebclrsn: I personally find it more explicit to set it to 0 in all failing cases

Pog3k: I also like the fact that we don't catch all errors

PR #463, opened 2024-03-05
Make release 0.27.1

sebclrsn:


sebastianpfischer (approved):

sebastianpfischer:
@sebclrsn , can you please bump up the pytest version?

sebclrsn:
> @sebclrsn , can you please bump up the pytest version?

I don't think that's feasible with poetry (updating only the lock file and for a single dependency) except if I explicitly set the minimum pytest version to 8.1

This PR adds compatibility with recent pytest versions, but doesn't enforce them :)

sebastianpfischer:
@sebclrsn , yes you can. We already pushed the lock file: https://github.com/eclipse/kiso-testing/blob/master/poetry.lock so just call `poetry update`  and push it. No?

PR #465, opened 2024-03-18
feat: add start stop can trace

yannpoupon:
Add the possibility to stop and start pcan trace and to specify name of the logfile wanted.

Since the user can specify a new path when he starts the pcan trace he can choose a different repository, I added a dictionary that has for key the directory path and as value a list of all the file_name for this repository to be able to rename all of them or to merge them depending on the logic wanted.

### DIFF
--- a/src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
+++ b/src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
@@ -18,7 +18,6 @@ Can Communication Channel using PCAN hardware
 .. currentmodule:: cc_pcan_can
 
 """
-
 import logging
 import os
 import shutil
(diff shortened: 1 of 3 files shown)


src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
```diff
@@ -167,7 +166,8 @@ def __init__(
         self.boottime_epoch = boottime_epoch
         self._initialize_trace()
         self.merge_trc_logs = merge_trc_logs
-
+        self.trc_file_names: dict[Path, str | None] = {}
```
sebclrsn, line 169: This should probably be made private

sebclrsn (approved):

sebastianpfischer (approved):

PR #466, opened 2024-03-25
feat: com aux able to receive timestamps

TedRio:


sebastianpfischer (approved):

src/pykiso/lib/auxiliaries/communication_auxiliary.py
```diff
@@ -167,10 +167,15 @@ def receive_message(
 
         msg = response.get("msg")
         remote_id = response.get("remote_id")
+        timestamp = response.get("timestamp")
 
         # stay with the old return type to not making a breaking change
-        if remote_id is not None:
+        if remote_id and timestamp:
+            return (msg, remote_id, timestamp)
```
sebclrsn, line 174: That's still a breaking change, or am I missing something?

sebclrsn: Please add an additional keyword argument `receive_timestamp: bool = False` so that no breaking change occurs

sebclrsn (commented):
Please add an entry to the what's new then we're good to go :)

src/pykiso/lib/auxiliaries/communication_auxiliary.py
```diff
@@ -136,6 +136,7 @@ def receive_message(
         self,
         blocking: bool = True,
         timeout_in_s: float = None,
+        receive_timestamp: bool = False,
     ) -> Optional[bytes]:
```
sebclrsn, line 140:
```suggestion
    ) -> Optional[bytes | Tuple[bytes, int] | Tuple[bytes, int, float]]:
```

src/pykiso/lib/auxiliaries/communication_auxiliary.py
```diff
@@ -136,6 +136,7 @@ def receive_message(
         self,
         blocking: bool = True,
         timeout_in_s: float = None,
+        receive_timestamp: bool = False,
```
sebclrsn, line 139: Parameter docstring is missing 🙈

PR #476, opened 2024-05-27
fix: stop rx task call before start rx task is called

yannpoupon:
Fix a rare case where the `join()` function could be called before the thread is started.


sebastianpfischer:
A rebase is first needed to compare with the updated ci checks

sebastianpfischer:
@yannpoupon , looks fine for me. Is there a way to verify it?

sebastianpfischer (approved):

sebastianpfischer:
I guess I can live with that ^^

PR #484, opened 2024-07-01
feat: add one trace file per test case or test run

yannpoupon:
Add the possibility for the user to get one trace can per TestCase or for every test run from the configuration of the `CCPCanCan` by adding one parameter.

### DIFF
--- a/docs/connectors/CChannels/cc_pcan_can.rst
+++ b/docs/connectors/CChannels/cc_pcan_can.rst
@@ -15,7 +15,7 @@ The logging can be deactivated by passing the parameter `logging_activated` to F
 
 The strategy for the creation of the trace file can also be modified to create trace file for every test run or
 for every testCase run, by adding in the configuration of the CCPCanCan the parameter strategy_trc_file
-that take two possible value : "test" or "testCase"
+that take two possible values : "testRun" or "testCase"
 
 .. code:: yaml
 
@@ -26,4 +26,4 @@ that take two possible value : "test" or "testCase"
         channel: "PCAN_USBBUS1"
         state: "ACTIVE"
         type: pykiso.lib.connectors.cc_pcan_can:CCPCanCan
-        strategy_trc_file: "test"
+        strategy_trc_file: "testRun"
--- a/docs/connectors/CChannels/cc_pcan_can.rst
+++ b/docs/connectors/CChannels/cc_pcan_can.rst
@@ -11,11 +11,11 @@ Trace file Strategy
 By default the CCPCanCan will create one trace file for all the test executed when the command `pykiso -c`
 is executed.
 
-The logging can deactivated by passing the parameter `logging_activated` to False in the configuration.
+The logging can be deactivated by passing the parameter `logging_activated` to False in the configuration of the connector.
 
-But the strategy for the creation of the trace file can be modified to create trace file for every test run or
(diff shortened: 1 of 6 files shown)


Pog3k (approved):

src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
```diff
@@ -540,6 +547,9 @@ def stop_pcan_trace(self):
         Stops the PCAN trace if it is currently running.
         :return: None
         """
+        if not self.logging_activated:
+            log.warning("Logging is not activated")
```
sebastianpfischer, line 551: Why is this a warning?

docs/connectors/CChannels/cc_pcan_can.rst
```diff
@@ -4,3 +4,26 @@ cc_pcan_can
...
+-------------------
+
+By default the CCPCanCan will create one trace file for all the test executed when the command `pykiso -c`
+is executed.
+
+The logging can be deactivated by passing the parameter `logging_activated` to False in the configuration of the connector.
+
+The strategy for the creation of the trace file can also be modified to create trace file for every test run or
+for every testCase run, by adding in the configuration of the CCPCanCan the parameter strategy_trc_file
+that take two possible value : "test" or "testCase"
```
sebastianpfischer, line 18: `test` is too generic for me, can we have something more precise like `testRun` or `TestSuite` ?

src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
```diff
@@ -559,10 +569,16 @@ def start_pcan_trace(self, trace_path: Optional[str] = None, trace_size: int = 1
...
+            return
         if self.trace_running:
             log.warning("Trace is already started")
             return
 
+        if self.raw_pcan_interface is None:
+            self._cc_open()
```
sebastianpfischer, line 580: The trace should be started only if the channel is open. We also have locks to protect the resources and this is here a bypass.

sebastianpfischer (changes requested):

src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
yannpoupon: I added this  `cc_open` because if you use the logic with the auxiliary with auto_start to True, then the connection will not be opened then I can't start the  pcan trace automatically.

src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
yannpoupon: It was because the user could call the function.

docs/connectors/CChannels/cc_pcan_can.rst
yannpoupon: I changed it to `testRun`

TedRio (approved):

src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
sebastianpfischer: But sorry, if this is a functionality that should be provided with the auto-start, it needs to become a parameter and also started there. As of today, this is a API the user calls and should stay like that.

Can you please provide a fix for that?

PR #489, opened 2024-07-19
feat: add context manager to receive can message

yannpoupon:
Add a context manager to be able to receive a message after a send operation in case the response is too fast.

### DIFF
--- a/src/pykiso/lib/auxiliaries/can_auxiliary/can_auxiliary.py
+++ b/src/pykiso/lib/auxiliaries/can_auxiliary/can_auxiliary.py
@@ -19,9 +19,14 @@ The goal of this module is to be able to receive and send signals which are defi
 
 .. currentmodule:: can_auxiliary
 """
+from __future__ import annotations
 
+import functools
 import logging
+import threading
 import time
+from contextlib import ContextDecorator
+from copy import deepcopy
 from queue import Empty, Queue
 from typing import Any, Optional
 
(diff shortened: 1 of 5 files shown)


src/pykiso/lib/auxiliaries/can_auxiliary/can_auxiliary.py
```diff
@@ -130,6 +161,13 @@ def get_last_signal(self, message_name: str, signal_name: str) -> Optional[Any]:
 
         return None
 
+    def get_collected_messages(self) -> list[CanMessage]:
+        """Get all the messages collected with the context manager
+
+        :return: a list with all the collected messages
+        """
+        return deepcopy(self._messages_collected)
```
sebastianpfischer, line 169: You probably should clear the `self._messages_collected` list

src/pykiso/lib/auxiliaries/can_auxiliary/can_auxiliary.py
```diff
@@ -52,6 +78,9 @@ def __init__(self, com: CChannel, dbc_file: str, **kwargs):
         path_to_dbc_file = dbc_file
         self.parser = CanMessageParser(path_to_dbc_file)
         self.can_messages = {}
+        self._collect_msg = threading.Event()
```
sebastianpfischer, line 81:
```suggestion
        # Variables to manage the collection of can messages with a context-manager
        self._collect_msg = threading.Event()
```

sebastianpfischer (changes requested):

src/pykiso/lib/auxiliaries/can_auxiliary/can_auxiliary.py
yannpoupon: Done

src/pykiso/lib/auxiliaries/can_auxiliary/can_auxiliary.py
yannpoupon: It is clear whenever the context manager is entered.

TedRio (approved):

src/pykiso/lib/auxiliaries/can_auxiliary/can_auxiliary.py
Pog3k: I think also its nicer so get_collected_messages can be called multiple times.

Pog3k (approved):

PR #490, opened 2024-07-23
feat: force pykiso to close if threads are blocking

Pog3k:
Solves the issue when users spawns threads and Pykiso keeps the python process open when user threads are unresolved.

### DIFF
--- a/src/pykiso/cli.py
+++ b/src/pykiso/cli.py
@@ -19,8 +19,11 @@ Integration Test Framework
 
 """
 import logging
+import os
 import pprint
 import sys
+import threading
+import time
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
@@ -33,6 +36,8 @@ from .logging_initializer import change_logger_class, initialize_logging
 from .test_coordinator import test_execution
 from .test_setup.config_registry import ConfigRegistry
 from .types import PathType
+
+UNRESOLVED_THREAD_TIMEOUT = 10
 
 
 def eval_user_tags(click_context: click.Context) -> Dict[str, List[str]]:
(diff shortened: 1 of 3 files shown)


TedRio (approved):

yannpoupon (approved):

src/pykiso/cli.py
```diff
@@ -84,6 +89,35 @@ def check_file_extension(click_context: click.Context, param: click.Parameter, p
     return paths
 
 
+def active_threads() -> list[str]:
+    """Get the names of all active threads except the main thread."""
```
sebastianpfischer, line 93: docstring not correct

src/pykiso/cli.py
```diff
@@ -84,6 +89,35 @@ def check_file_extension(click_context: click.Context, param: click.Parameter, p
...
+    return [thread.name for thread in threading.enumerate()][1:]
+
+
+def check_and_handle_unresolved_threads(log: logging.Logger, timeout: int = 10) -> None:
```
sebastianpfischer, line 97: docstring also not correct

src/pykiso/cli.py
```diff
@@ -84,6 +89,35 @@ def check_file_extension(click_context: click.Context, param: click.Parameter, p
...
+        time.sleep(timeout)
+        if threading.active_count() > 1:
+            log.fatal(
+                f"Unresolved threads {', '.join(active_threads())} are still running after {timeout} seconds. Force pykiso to Exit."
+            )
+            os._exit(test_execution.ExitCode.UNRESOLVED_THREADS)
+        else:
+            log.warning("Unresolved threads has been properly shut down. Normal exit.")
+
+
```
sebastianpfischer, line 120: This code should not be part of the cli! This should be in https://github.com/eclipse/kiso-testing/blob/master/src/pykiso/test_setup/dynamic_loader.py#L333! Please provide a fix!

PR #493, opened 2024-08-06
Fix/add ccpcan init values

TedRio:


src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
```diff
@@ -575,7 +578,7 @@ def start_pcan_trace(self, trace_path: Optional[str] = None, trace_size: int = 1
             log.warning("Trace is already started")
             return
 
-        if self.raw_pcan_interface is None:
+        if self.auto_start and self.raw_pcan_interface is None:
```
sebastianpfischer, line 581: I would have assumed that setting the `auto-start` would eventually call `cc_open` at a certain point. If the parameters on the init are extended with a bool to start the trace, we can have it started after cc_open was called.
We, then, will be able to delete the `self._cc_open()` here

sebastianpfischer: Set a meeting with me to discuss that

TedRio: Should be fix with new commit

sebastianpfischer (approved):

PR #495, opened 2024-08-13
fix: remove incompability between instrument aux and serial connector

TedRio:


src/pykiso/lib/connectors/cc_serial.py
```diff
@@ -198,6 +198,9 @@ def _cc_send(self, msg: ByteString, timeout: float = None, **kwargs) -> None:
             self.current_write_timeout = timeout
             self.serial.write_timeout = timeout
 
+        if isinstance(msg, str):
```
sebastianpfischer, line 201: Maybe add a comment on top to explain this line? The coverage says you are not verifying it.

Pog3k (approved):

sebastianpfischer (approved):

PR #504, opened 2024-09-21
feat: configuration and auxiliaries can be started and used in any py…

sebastianpfischer:
Separate pykiso-testing-framework from the simulation part (with auxiliaries).

I need support for two things:
* Correct location for placing the tests
* General verification that the changes do not impact negatively the existing code


examples/next_pykiso2/pykiso_as_simulator/serial_simulation.py
```diff
@@ -0,0 +1,30 @@
+from pathlib import Path
+
+import pykiso
+
+pykiso.load_config(Path(__file__).parent.resolve() / "serial.yaml")
+
+
+from pykiso.lib.auxiliaries.communication_auxiliary import CommunicationAuxiliary
+
+sender = CommunicationAuxiliary.get_instance('com_aux_sender')
```
Pog3k, line 10: Cool would be a factory.
I mean the info that the aux is of type  `CommunicationAuxiliary` is already in the yaml.

kiso_conf = pykiso.load_config(Path(__file__).parent.resolve() / "serial.yaml")
kiso_conf.get_instance('com_aux_sender')

maybe for the future ?!

Pog3k (approved):

examples/next_pykiso2/pykiso_as_simulator/serial_simulation.py
sebastianpfischer: One step after the other ^^

sebastianpfischer: I'm not sure on how to resolve the the typing resolution for that

src/pykiso/auxiliary.py
```diff
@@ -212,7 +224,8 @@ def _start_rx_task(self) -> None:
         with self.rx_lock:
             task_name = f"{self.name}_rx"
             log.internal_debug("start reception task %s", task_name)
-            self.rx_thread = threading.Thread(name=task_name, target=self._reception_task)
+            # Any created thread should disappear after main-thread exit
+            self.rx_thread = threading.Thread(name=task_name, target=self._reception_task, daemon=True)
```
sebclrsn, line 228: We should definitely test the different auxiliaries thoughtfully before merging this change, switching to daemon threads could have quite a lot of unexpected consequences (and if the auxiliries are "killed" at the end, we'll probably lose some insights on the correctness of our thread implementations)

src/pykiso/auxiliary.py
```diff
@@ -257,6 +270,21 @@ def stop(self) -> bool:
         """
         return self.delete_instance()
 
+    def __enter__(self) -> Self:
+        """Context manager entry point"""
+        if self.start():
+            return self
+        else:
+            raise RuntimeError(f"Failed to start auxiliary {self.name}")
```
sebclrsn, line 278: Didn't we have dedicated exceptions instead?

tests/test_dt_auxiliary_interface.py
```diff
@@ -229,6 +230,38 @@ def test_stop(mocker, aux_inst):
 
     assert state is True
 
+def test_aux_as_context_manager(mocker, aux_inst):
+    create_inst_mock: Mock = mocker.patch.object(aux_inst, "create_instance", return_value=True)
+    delete_inst_mock: Mock = mocker.patch.object(aux_inst, "delete_instance", return_value=True)
+
+    with aux_inst as _:
```
sebclrsn, line 237:
```suggestion
    with aux_inst:
```

sebclrsn (approved):
Apart from the change to daemonic threads at the core of the auxiliaries, I don't see a major/breaking change here. It's nice to have that feature!

src/pykiso/auxiliary.py
sebastianpfischer: @sebclrsn 
When I start the context manager with nothing in, it sometimes hang. From what I have seen, the proxy does not exit correctly. I'm not sure on how to resolve this issue. Maybe you can take a look at it?

PR #512, opened 2024-11-04
fix: the acroname discoverAndConnect method does not accept None anymore…

flo117:
... use 0 instead.

In the cli the exit code will be passed through since the tests could still have passed. Main issue is the lib lgpio, which opens a thread, as a daemon, and not cl

### DIFF
--- a/src/pykiso/cli.py
+++ b/src/pykiso/cli.py
@@ -89,13 +89,19 @@ def check_file_extension(click_context: click.Context, param: click.Parameter, p
     return paths
 
 
-def active_threads() -> list[str]:
+def active_threads() -> list[list[str]]:
     """Get the names of all active threads except the main thread."""
-    return [thread.name for thread in threading.enumerate()][1:]
+    frames = sys._current_frames()
+    return [[thread.name, str(frames.get(thread.ident))] for thread in threading.enumerate()][1:]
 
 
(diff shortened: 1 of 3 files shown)


Pog3k (approved):

src/pykiso/cli.py
```diff
@@ -89,13 +89,19 @@ def check_file_extension(click_context: click.Context, param: click.Parameter, p
     return paths
 
 
-def active_threads() -> list[str]:
+def active_threads() -> list[list[str]]:
```
Pog3k, line 92: Is it not ?
```suggestion
def active_threads() -> list[tuple[str, str]]:
```

src/pykiso/cli.py
```diff
@@ -89,13 +89,19 @@ def check_file_extension(click_context: click.Context, param: click.Parameter, p
...
     """Get the names of all active threads except the main thread."""
```
Pog3k, line 93: Now you also return where the code is running. so the docstrings needs to be adapted

src/pykiso/cli.py
flo117: changed the return value and typing to list[tuple[str, str]]

sebastianpfischer (approved):

src/pykiso/cli.py
flo117: done

PR #516, opened 2024-11-14
feature: unify path specifier for can connectors. Fix open thread iss…

Pog3k:
Seems that circular dependencies prevent the del dunder to be called.
Converted it to a normal stop method.

Also change log_path to trace_path so users can switch between cc pcan and cc socketcan without the need to adapt those parameters.


src/pykiso/lib/connectors/cc_socket_can/cc_socket_can.py
```diff
@@ -54,15 +54,15 @@ class CCSocketCan(CChannel):
     def __init__(
         self,
         channel: str = "vcan0",
-        remote_id: int = None,
+        remote_id: Optional[int] = None,
```
flo117, line 57: int | None

src/pykiso/lib/connectors/cc_socket_can/cc_socket_can.py
```diff
@@ -75,7 +75,7 @@ def __init__(
         :param is_extended_id: this flag controls the size of the arbitration_id field
         :param receive_own_messages: if set transmitted messages will be received
         :param logging_activated: boolean used to enable logfile creation
-        :param log_path: trace directory path (absolute or relative)
+        :param log_path: trace directory path (absolute or relative), if you set a trc file name, auto timestamp and log_name will be ignored
```
yannpoupon, line 78:
```suggestion
        :param trace_path: trace directory path (absolute or relative), if you set a trc file name, auto timestamp and log_name will be ignored
```

yannpoupon (approved):

src/pykiso/lib/connectors/cc_socket_can/cc_socket_can.py
yannpoupon: This functionality is available since python 3.10 but pykiso supports python 3.9 : https://github.com/eclipse-kiso-testing/kiso-testing/blob/master/pyproject.toml#L35

src/pykiso/lib/connectors/cc_socket_can/cc_socket_can.py
```diff
@@ -89,29 +89,40 @@ def __init__(
...
-            dir_path = (
-                (Path() / self.log_path).resolve() if not Path(self.log_path).is_absolute() else Path(self.log_path)
-            )
-            # if no specific logging file name is given take the default one
-            self.log_name = (
-                time.strftime(f"%Y-%m-%d_%H-%M-%S_{self.log_name}.trc")
-                if self.log_name is not None
-                else time.strftime("%Y-%m-%d_%H-%M-%S_CanLog.trc")
+            self.trace_path = (
+                (Path().resolve() / self.trace_path).resolve()
```
flo117, line 106: i think the outer resolve has no effect

flo117: could you just do Path(self.trace_path).resolve()

src/pykiso/lib/connectors/cc_socket_can/cc_socket_can.py
Pog3k: fixed

src/pykiso/lib/connectors/cc_socket_can/cc_socket_can.py
Pog3k: yupp

PR #517, opened 2024-11-14
Fix resuming CCTcpip

thc3si:
When an auxiliary such as `InstrumentControlAuxiliary` is `suspend`ed then `resume`d, its channel has its methods `_cc_close`, then `_cc_open` called. 

The `CCTcpip` connector currently doesn't support that: once `_cc_close` has been called, `_cc_open` cannot be called anymore. This is because `socket.socket` does not support reopening a closed socket. 

The changes here create the socket in `_cc_open`, allowing multiple `_cc_open` / `_cc_close` cycles. 

Relevant test (`tests/test_cc_tcp_ip.py`) has been updated, because it depended on using various methods of `CCTcpip` without first calling `_cc_open`. This use case should not happen in the real world, and doesn't work anymore now that the socket is only created in `_cc_open`.

### DIFF
--- a/src/pykiso/lib/connectors/cc_tcp_ip.py
+++ b/src/pykiso/lib/connectors/cc_tcp_ip.py
@@ -41,7 +41,7 @@ class CCTcpip(CChannel):
         super().__init__(**kwargs)
         self.dest_ip = dest_ip
         self.dest_port = int(dest_port)
-        self.socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.socket: socket.socket | None = None
         self.max_msg_size = max_msg_size
         # Set a timeout to send the signal to the GIL to change thread.
         # In case of a multi-threading system, all tasks will be called one after the other.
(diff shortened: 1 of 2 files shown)


src/pykiso/lib/connectors/cc_tcp_ip.py
```diff
@@ -76,6 +83,9 @@ def _cc_receive(self, timeout=0.01) -> Dict[str, Optional[bytes]]:
 
         :return: Message if successful, otherwise none
         """
+        if not self.socket:
+            raise TypeError("Channel must be opened before messages can be received")
```
Pog3k, line 87: same here

src/pykiso/lib/connectors/cc_tcp_ip.py
```diff
@@ -41,29 +41,36 @@ def __init__(self, dest_ip: str, dest_port: int, max_msg_size: int = 256, **kwar
...
+        self.socket: socket.socket | None = None
         self.max_msg_size = max_msg_size
         # Set a timeout to send the signal to the GIL to change thread.
         # In case of a multi-threading system, all tasks will be called one after the other.
         self.timeout = 1e-6
 
     def _cc_open(self) -> None:
         """Connect to socket with configured port and IP address."""
+        if self.socket:
+            return
```
flo117, line 53: maybe add a warning

flo117 (approved):

Pog3k (approved):

yannpoupon (approved):

PR #529, opened 2025-01-30
Feature/add profiling

Pog3k:
Adding decorator for profiling a test


Pog3k:
Seems that not everything is traced from the beginning. Threads are missing. See this PR as draft PR

Pog3k:
> Seems that not everything is traced from the beginning. Threads are missing. See this PR as draft PR

import profiling in cli.py fixed the issue

docs/advanced_usage/profiling.rst
```diff
@@ -0,0 +1,99 @@
...
+
+```python
+from pykiso.profiling import profile
+
+@profile
+def test_method():
+    pass
+
+@profile
+def test_method("result.json"):
```
yannpoupon, line 30: "result.json" should be in the decorator argument.

yannpoupon (approved):

docs/advanced_usage/profiling.rst
Pog3k: fixed

PR #533, opened 2025-02-19
fix: auxiliary transmit task killed when creating the instance

yannpoupon:


yannpoupon:
Issue with unittest do not merge

PR #535, opened 2025-03-19
Feature/parameterized tests with xray poc

DianeMornas:
The goal of this POC is to change the way to upload test results to xray, before 2 steps were required first step run the test and generate the jjunit xml file at the same time then second step convert this junit into a junit compatible with xray to upload the test results. The parameterized were not supported. This way only worked with xray test ticket with a Generic type and not with Manual, Automated and automated Cucumber types.

The user still uses a decorator with the test_key ID of the xray test ticket 
- if the pykiso test does not use parameterized
- only test execution ticket will be create in xray for all the test function and linked to the xray test tickets
- if the pykiso test use parameterized
- one test execution ticket will be created in xray per test function and per parameters

HTML step report generation has also been updated and support parameterized tests.

### DIFF
--- a/docs/tools/xray.rst
+++ b/docs/tools/xray.rst
@@ -4,7 +4,7 @@
 Export results to Xray
 ======================
 
-The ``xray`` CLI utility takes your Pykiso Junit report and export them on `Xray <https://xray.cloud.getxray.app/>`__.
+The ``xray`` CLI utility takes your Pykiso Junit test results report and export them on `Xray <https://xray.cloud.getxray.app/>`__.
 
 Upload your results
 -------------------
@@ -12,15 +12,13 @@ To upload your results to Xray users have to follow the command :
 
 .. code:: bash
 
-    xray --user USER_ID --password MY_API_KEY --url https://xray.cloud.getxray.app/ upload  --path-results path/to/reports/folder --test-execution-id "BDU3-12345"
-    --project-key KEY
+    xray --user USER_ID --password MY_API_KEY --url "https://xray.cloud.getxray.app/" upload --path-results path/reports/folder --test-execution-id
 
 Options:
   --user TEXT                   Xray user id  [required]
   --password TEXT               Valid Xray API key (if not given ask at command prompt
                                 level)  [optional]
   --url TEXT                    URL of Xray server  [required]
-  --project-key TEXT            Key of the project  [required]
   --path-results PATH           Full path to the folder containing the JUNIT reports
                                 [required]
   --test-execution-id TEXT      Xray test execution ticket id's use to import the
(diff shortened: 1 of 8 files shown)


sebastianpfischer (changes requested):
Hi @DianeMornas , there are core design issues with this pull-request. Could you please get in touch with @Pog3k so that he setup a meeting with all of us to discuss it. Could you also please create a ticket associated to this pull-request and the needs?

Thank you in advance

src/pykiso/test_coordinator/test_execution.py
sebastianpfischer, line 1: The generation of the needed information should be created after the test was executed. If information needs to be fetchted during execution, most of the content should be part of the test_result module. 

If we need to record now each line executed from a specific script, we need to take a look at it as well.

So I expect 2 parts: one in the test_result and maybe another here but more in a way we store information in a generic way that can per retrieve from the test_result itself

src/pykiso/cli.py
```diff
@@ -231,6 +265,20 @@ def parse_args(self, ctx, args):
...
+    required=False,
+    help="Upload the test results to Xray. Requires the username, password and url as arguments.",
+)
+@click.option(
+    "--test-execution-id",
+    nargs=1,
+    type=click.STRING,
+    required=False,
+    help="Test execution ID of Xray where to upload the test results. By default, a new test execution ticket is created except if the ID is given.",
+)
```
sebastianpfischer, line 281: xray should not become an interface of pykiso. The upload should still be done via the tool.

sebastianpfischer (changes requested):
Like said above, we need to have a design discussion. In addition, tests,  documentation and examples are missing

Pog3k:
What I have seen in this POC:

assert_step_report has been improved to collect additional test information, which will be used for the step report HTML and to provide all necessary details for the Xray report.

In test_execution.py, additional functions related to Xray reports have been added. I think it would be cleaner if we moved these new functions to a separate file/module dedicated to Xray functions, perhaps something like tool/xray/xray_report.py.

In assert_step_report.py, we are mixing step report functions with Xray again. I think it makes sense to move generate_step_report_xray to tool/xray/xray_report.py. The refactored step report logic is fine because both require the same dictionary, ALL_STEP_REPORT.

I believe we need to modify the Pykiso interface because the information in the JUnit report is insufficient to capture all necessary details. See the improvements in assert_step_report. Somehow, we need make the informations stored ALL_STEP_REPORT available. Since this is quite generic, we could also introduce a flag like --kiso-report, which dumps the pickled dictionary or something similar.

For the upload process, we could still use the external CLI, which has the benefit of allowing results to be uploaded at any time. This way, tests don’t need to be re-run if something fails.

I'll be back on Monday to discuss this further. However, feel free to leave any comments in advance.

sebastianpfischer:
> What I have seen in this POC:
> 
> assert_step_report has been improved to collect additional test information, which will be used for the step report HTML and to provide all necessary details for the Xray report.
> 
> In test_execution.py, additional functions related to Xray reports have been added. I think it would be cleaner if we moved these new functions to a separate file/module dedicated to Xray functions, perhaps something like tool/xray/xray_report.py.
> 
> In assert_step_report.py, we are mixing step report functions with Xray again. I think it makes sense to move generate_step_report_xray to tool/xray/xray_report.py. The refactored step report logic is fine because both require the same dictionary, ALL_STEP_REPORT.
> 
> I believe we need to modify the Pykiso interface because the information in the JUnit report is insufficient to capture all necessary details. See the improvements in assert_step_report. Somehow, we need make the informations stored ALL_STEP_REPORT available. Since this is quite generic, we could also introduce a flag like --kiso-report, which dumps the pickled dictionary or something similar.
> 
> For the upload process, we could still use the external CLI, which has the benefit of allowing results to be uploaded at any time. This way, tests don’t need to be re-run if something fails.
> 
> I'll be back on Monday to discuss this further. However, feel free to leave any comments in advance.

It makes sense. But we should take a deeper look at the report mechanism all together. I believe this may need some major design decision.

Pog3k (approved):

PR #540, opened 2025-04-06
fix: different ci/cd related indicators

sebastianpfischer:


pyproject.toml
```diff
@@ -30,7 +30,7 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.9"
+python = "^3.12"
```
Pog3k, line 33: maybe we have still 3.10 users.. I would like to anounce it first internally..
Im sick this week maybe we could update it in the next release. 
If you want to update 3.10 otherwise dont touch the line pls (:

Pog3k (changes requested):
Current Python limit 3.10

Pog3k (approved):

PR #542, opened 2025-04-17
Feature/ html step report supports parameterized tests

DianeMornas:
Improve the html step reports by supporting the parameterized tests

### DIFF
--- a/src/pykiso/test_result/assert_step_report.py
+++ b/src/pykiso/test_result/assert_step_report.py
@@ -60,6 +60,8 @@ REPORT_KEYS = [
     "expected_result",
     "actual_result",
     "succeed",
+    "failure_log",
+    "description",
 ]
 DEFAULT_TEST_METHOD = ["setup", "teardown", "handle_interaction"]
 # Parent method being reported ; Ignore sub call (assert in an assert)
@@ -233,7 +235,7 @@ def _prepare_report(test: unittest.case.TestCase, test_name: str) -> None:
     # Create the current test step storage
     if not ALL_STEP_REPORT[test_class_name]["test_list"].get(test_name):
         test_method = getattr(test, test_name, None)
-        test_description = test_method.__doc__ or ""
+        test_description = (test_method.__doc__ or test._testMethodDoc or "").split(" [")[0]
         ALL_STEP_REPORT[test_class_name]["test_list"][test_name]: Dict[
             str,
             Union[str, List[List[Dict[str, Union[str, bool]]]], List[List[str]]],
@@ -251,11 +253,24 @@ def _add_step(
     var_name: str,
     expected: typing.Any,
     received: typing.Any,
+    failure_log: str,
 ):
     global ALL_STEP_REPORT, REPORT_KEYS
 
     ALL_STEP_REPORT[test_class_name]["test_list"][test_name]["steps"][-1].append(
-        dict(zip(REPORT_KEYS, [message, var_name, expected, received, True]))
+        dict(
+            zip(
+                REPORT_KEYS,
+                [
+                    message,
+                    var_name,
+                    expected,
+                    received,
+                    True,
+                    failure_log,
+                ],
+            )
+        )
     )
 
 
@@ -348,7 +363,7 @@ def assert_decorator(assert_method: types.MethodType):
             signature = inspect.signature(assert_method)
             arguments = signature.bind(*args, **kwargs).arguments
             test_name = test_case_inst.step_report.current_table or test_name
-            # 1. Gather message, var_name, expected, received
+            # 1. Gather message, var_name, expected, received, failure_log
             # 1.1 Get message. default value: ""
             if test_case_inst.step_report.message:
                 message = test_case_inst.step_report.message
@@ -388,6 +403,9 @@ def assert_decorator(assert_method: types.MethodType):
             # 1.4. Get Expected value
             expected = _get_expected(assert_name, arguments)
 
+            # 1.5. Get the failure logs
+            failure_log = traceback.format_exc() if test_case_inst.failureException else "None"
+
             # 2. Update report data
             # 2.1 Ensure report ready for update
             _prepare_report(test_case_inst, test_name)
@@ -400,6 +418,7 @@ def assert_decorator(assert_method: types.MethodType):
                 var_name,
                 expected,
                 received,
+                failure_log,
             )
 
         except Exception as e:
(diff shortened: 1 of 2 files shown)


src/pykiso/test_result/assert_step_report.py
```diff
@@ -501,8 +525,16 @@ def generate_step_report(
             ALL_STEP_REPORT[class_name]["time_result"]["End Time"] = stop_time
             ALL_STEP_REPORT[class_name]["time_result"]["Elapsed Time"] = round(elapsed_time, 2)
             if test_case in test_result.errors:
-                ALL_STEP_REPORT[class_name]["test_list"][test_method_name]["unexpected_errors"][-1].append(test_case[1])
-                ALL_STEP_REPORT[class_name]["succeed"] = False
+                test_names = [
+                    key
+                    for key in ALL_STEP_REPORT[class_name]["test_list"].keys()
+                    if (key.startswith("test_") and key != "test_run")
+                ]
```
sebastianpfischer, line 532: This feels like you are filtering the tests, I'm not sure this is ok.

tests/test_assert_step_report.py
```diff
@@ -102,7 +102,8 @@ def test_assert_decorator_no_message(mocker, test_case):
         "",
         "data_to_test",
         "True",
-        data_to_test,
+        True,
```
sebastianpfischer, line 105: You should probably keep `data_to_test` here

tests/test_assert_step_report.py
```diff
@@ -120,7 +121,8 @@ def test_assert_decorator_step_report_message(mocker, test_case):
         "Dummy message",
         "data_to_test",
         "True",
-        data_to_test,
+        True,
```
sebastianpfischer, line 124: Same for all the expected outputs

tests/test_test_case.py
```diff
@@ -393,7 +393,7 @@ class DummyClass(test_case.BasicTest):
         pass
 
     test_case_inst = DummyClass(1, 1, None, {"Component1": ["Req1", "Req2"]}, None, None, None, None)
-    assert test_case_inst.properties is None
+    assert test_case_inst.properties == None
```
sebastianpfischer, line 396: What was wrong here?

sebastianpfischer (changes requested):
I still need to deepdive into your tests

src/pykiso/test_result/assert_step_report.py
```diff
@@ -233,7 +235,7 @@ def _prepare_report(test: unittest.case.TestCase, test_name: str) -> None:
     # Create the current test step storage
     if not ALL_STEP_REPORT[test_class_name]["test_list"].get(test_name):
         test_method = getattr(test, test_name, None)
-        test_description = test_method.__doc__ or ""
+        test_description = (test_method.__doc__ or test._testMethodDoc or "").split(" [")[0]
```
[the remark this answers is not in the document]
DianeMornas: Yes sure,
here a simple parameterized test with 2 test cases: (True), (False):

```
@pykiso.define_test_parameters(suite_id=1, case_id=1, aux_list=[aux1])
class MyTest1(pykiso.BasicTest):
    """MyTest1"""
    @parameterized.expand([(True), (False)])
    def test_1(self, is_true):
        """Test run 1: check the assert true"""
        self.assertTrue(is_true, f"{is_true} should be True")
```

 ```test_method.__doc__``` is the documentation of the current test_method. With parameterized tests, there is as much documentation as test cases. 
For the first test case we get: 'Test run 1: check the assert true [with *args=(True,)]', and for the second test case we get: 'Test run 1: check the assert true [with *args=(False,)]'. 

In the html step report will appear the documentation of the last set of parameters of the test parameterized, so before the change with ```test_method.__doc__ or test._testMethodDoc or ""```:
'Test run 1: check the assert true [with *args=(True,)]'

after the change with ```(test_method.__doc__ or test._testMethodDoc or "").split(" [")[0]```
'Test run 1: check the assert true'

With this change we avoid to display the last parameter of the last test case. 
The test documentation remains generic.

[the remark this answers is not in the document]
DianeMornas: I can not add screenshot, I do not understand why...

src/pykiso/test_result/assert_step_report.py
DianeMornas: Yes, here, I am filtering the test. It is a mistake, I have to check why we do not upload the results of test_run, setUp and tearDown ... I will fix it
These lines worked without parameterized tests. But when use the cli with parameterized tests there is an issue.
For example with this simple parameterized test:

```
@pykiso.define_test_parameters(suite_id=1, case_id=1, aux_list=[aux1])
class MyTest1(pykiso.BasicTest):
    """MyTest1"""
    @parameterized.expand([(True), (False)])
    def test_1(self, is_true):
        """Test run 1: check the assert true"""
        self.assertTrue(is_true, f"{is_true} should be True")
```

But here, it seems complicated because, there was an issue with parameterized test.
There are 2 test cases (True), (False) with 2 test_method_name values:
- test_1_0
- test_1_1

For the first test case,
```ALL_STEP_REPORT[class_name]["test_list"]``` contains

```OrderedDict([('test_1', {'description': 'Test run 1: check the assert true', 'steps': [[{'message': 'True should be True', 'var_name': 'is_true', 'expected_result': 'True', 'actual_result': True, 'succeed': True, 'failure_log': 'NoneType: None\n'}, {'message': 'False should be True', 'var_name': 'is_true', 'expected_result': 'True', 'actual_result': False, 'succeed': False, 'failure_log': 'Traceback (most recent call last):\n  File "C:\\Users\\MOI4RT\\Documents\\workspace_ITF_projects\\Open_source\\kiso-testing\\src\\pykiso\\test_result\\assert_step_report.py", line 431, in func_wrapper\n    return assert_method(*args, **kwargs)\n  File "C:\\Program Files\\Python310\\lib\\unittest\\case.py", line 687, in assertTrue\n    raise self.failureException(msg)\nAssertionError: False is not true : False should be True\n'}]], 'unexpected_errors': [[]]})])
```

So 
ALL_STEP_REPORT[class_name]["test_list"][test_method_name] raises a Key Error: ``` KeyError: 'test_1_0'```
That is why, I check the test method name
```name = test_name if test_name != test_method_name else test_method_name```

DianeMornas: I will check, why we do not upload the results of test_run, setUp and tearDown ... There is something weird here... need more investigations

tests/test_assert_step_report.py
DianeMornas: I will check, it is weird

tests/test_test_case.py
DianeMornas: I will check, it is weird

DianeMornas:
@sebastianpfischer , hello, I applied you review suggestions. I left some comments. Is the PR ok for you?

src/pykiso/test_result/assert_step_report.py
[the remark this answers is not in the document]
Pog3k: has been internally showcases and it works..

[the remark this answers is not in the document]
sebastianpfischer: @DianeMornas , actually, what I wanted was a comment in the code to understand why this change ^^

PR #557, opened 2025-07-24
Feat/log per parameter

yannpoupon:
Add the possibility to separate the log file in multiple log file : 
- One for every test case
- One for every test run

It will always create one log file that will hold all the test.

### DIFF
--- a/src/pykiso/test_result/multi_result.py
+++ b/src/pykiso/test_result/multi_result.py
@@ -20,17 +20,24 @@ Multi test result
 """
 from __future__ import annotations
 
+import logging
+import unittest
+from datetime import datetime
 from inspect import signature
-from typing import Any, List, Optional, TextIO, Union
-from unittest import TestResult
+from pathlib import Path
+from typing import Any, List, Literal, Optional, Union
+from unittest import TestResult, util
 from unittest.case import _SubTest
 
 from pykiso.types import ExcInfoType
 
+from ..logging_initializer import get_internal_level, get_logging_options
 from ..test_coordinator.test_case import BasicTest
 from ..test_coordinator.test_suite import BaseTestSuite
 from .text_result import BannerTestResult
 from .xml_result import XmlTestResult
+
+test_runner_instance: Optional[unittest.TextTestRunner] = None
 
 
 class MultiTestResult:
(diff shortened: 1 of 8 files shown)


src/pykiso/test_result/multi_result.py
```diff
@@ -99,9 +116,46 @@ def startTest(self, test: Union[BasicTest, BaseTestSuite]) -> None:
...
+        log_options = get_logging_options()
+
+        if test.__class__ in self._list_test_results and self._log_file_strategy == "testCase":
+            log_path = self.current_log_file
+        else:
+            log_file_name = util.strclass(test.__class__).replace(".", "_").replace("-", "_")
+            if self._log_file_strategy == "testRun":
+                log_file_name += "_" + test._testMethodName
+
+            log_file_name += f"_{datetime.today().strftime('%Y%d%m%H%M%S')}.log"
```
sebastianpfischer, line 141: Why are you not reusing `self.current_log_file` as base for the split?

src/pykiso/test_result/multi_result.py
```diff
@@ -99,9 +116,46 @@ def startTest(self, test: Union[BasicTest, BaseTestSuite]) -> None:
...
+
+        # Setup the file handler for logging
+        log_format = logging.Formatter("%(asctime)s [%(levelname)s] %(module)s:%(lineno)d: %(message)s")
+        root_logger = logging.getLogger()
+        file_handler = logging.FileHandler(log_path, "a+")
+        file_handler.setFormatter(log_format)
+        file_handler.name = "strategy_log_file_handler"
+        # always include internal logs in log files
+        file_handler.setLevel(get_internal_level(log_options.log_level))
+        root_logger.addHandler(file_handler)
```
sebastianpfischer, line 152: Why do you need to set again a file-handler?

sebastianpfischer:
BTW, unittest fails

src/pykiso/test_result/multi_result.py
yannpoupon: It was requested by @Pog3k that we keep the original log file that holds all the test log (so the default file handler), and we create a new file when the strategy is active so this file handler.

src/pykiso/test_result/multi_result.py
yannpoupon: It has been changed to use the self.current_log_file.

PR #562, opened 2025-08-11
fix: trace file name too long

yannpoupon:
Fix a case where the file name is too long to be renamed.

image

### DIFF
--- a/src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
+++ b/src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
@@ -537,7 +537,18 @@ class CCPCanCan(CChannel):
             list_of_traces = sorted(trace_path.glob("*.trc"), key=os.path.getmtime)[-len(trace_file_names) :]
             for index, file_name in enumerate(trace_file_names):
                 if file_name is not None:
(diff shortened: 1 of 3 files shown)


Pog3k (approved):

src/pykiso/lib/connectors/cc_pcan_can/cc_pcan_can.py
```diff
@@ -537,7 +537,18 @@ def _rename_trc(self):
...
+                        list_of_traces[index].rename(trace_path / file_name)
+                    except OSError as e:
+                        if len(file_name) > 255:  # Most filesystems have 255 char limit
+                            # Truncate filename while preserving extension
+                            name, ext = os.path.splitext(file_name)
+                            truncated_name = name[len(ext) - 255 :] + ext
+                            log.warning("Filename too long, truncating '%s' to '%s'", file_name, truncated_name)
+                            list_of_traces[index].rename(trace_path / truncated_name)
+                        else:
+                            log.error("Failed to rename trace file: %s", e)
```
Pog3k, line 550: dont we use log exceoption normaly ?

yannpoupon: The error is raised after so the log exception is not needed

